PR #3, opened 2017-05-05
Copter: respect CONDITION_YAW during landing

jschall:


ArduCopter/control_auto.cpp
```diff
@@ -387,7 +387,9 @@ void Copter::auto_land_start(const Vector3f& destination)
     }
 
     // initialise yaw
-    set_auto_yaw_mode(AUTO_YAW_HOLD);
+    if (auto_yaw_mode != AUTO_YAW_LOOK_AT_HEADING) {
```
presmini: just for my understanding, the AUTO_YAW_LOOK_AT_HEADING value...is that an AP_PARAM value?

PR #8, opened 2017-06-08
add support for FTS STATUS2 message

jschall:


jschall:
I've re-verified FTS watchdog deploy by programming Pixhawk to stop sending watchdog resets after a period of time, and I've re-verified that commanded deployments work by triggering a tilt angle error deployment. I've verified that all of the DEBUG messages get sent and have the correct information, and I've verified that the FTSS and FTS2 dataflash messages get logged at the correct rate.

PR #10, opened 2017-06-22
AP_Parachute: fix bug causing failure to deploy with 3-position switch

jschall:


jschall:
We have decided not to use the 3-position switch, so I am closing this.

PR #13, opened 2017-07-10
add arming check for matternet FTS

jschall:


jschall:
I just committed a patch that adds fuse fault to this check.

PR #22, opened 2018-03-23
AP_Notify: don't play any tones while armed

jschall:


libraries/AP_Notify/ToneAlarm_PX4.cpp
```diff
@@ -114,6 +114,11 @@ bool ToneAlarm_PX4::init()
 // play_tune - play one of the pre-defined tunes
 void ToneAlarm_PX4::play_tone(const uint8_t tone_index)
 {
+    // Just don't play tones while armed
```
MitchKoch: Consider indicating why in the comment. Perhaps,
Suppresses tones while armed as catastrophic interference with IMU has been observed with some tones.

PR #23, opened 2018-03-24
Fixes to mode switch, radio failsafe

jschall:


ArduCopter/flight_mode.cpp
```diff
@@ -24,6 +24,10 @@ bool Copter::set_mode(control_mode_t mode, mode_reason_t reason)
         return true;
     }
 
+    if ((control_mode == STABILIZE || control_mode == ALT_HOLD || control_mode == LOITER) && reason == MODE_REASON_GCS_COMMAND) {
```
derekxm: what is reason? When is reason MODE_REASON_GCS_COMMAND? This prevent Mavnet changing mode when vehicle is controlled by a pilot?

jschall:
Yep.

PR #26, opened 2018-05-02
Copter: fix coding error in radio failsafe

jschall:


jschall:
Paolo reviewed in person and verbally accepted

PR #28, opened 2018-05-24
Copter: prevent descent while centering for precision landing

jschall:


ArduCopter/control_land.cpp
```diff
@@ -178,8 +178,12 @@ void Copter::land_run_vertical_control(bool pause_descent)
 #endif
 
     // compute desired velocity
+    const float precland_acceptable_error_high = 100.0f;
+    const float precland_slow_descent_speed = 20.0f;
+    const float precland_slowdown_height = 100.0f;
+    const float precland_commit_height = 35.0f;
```
presmini: should all of these be configurable via parameter?

jschall: Probably. Adding parameters is a bit painful - we risk eventually running into conflicting indices with upstream parameter additions.

PR #36, opened 2018-11-14
Copter: run compass_checks in arming checks

jschall:
This should run the field strength check continuously up until the vehicle arms.


ArduCopter/AP_Arming.cpp
```diff
@@ -574,6 +574,10 @@ bool AP_Arming_Copter::arm_checks(bool display_failure, bool arming_from_gcs)
         }
         return false;
     }
+
```
Aksellence, line 577: Seems pretty straightforward.

Aksellence (approved):

jschall:
merged

Aksellence (commented):
Good

PR #37, opened 2018-11-14
Backport fixes to attitude controller re-initialization

jschall:
This is intended to fix the case where the vehicle yaws on the pad during takeoff.

### DIFF
--- a/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
+++ b/libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
@@ -114,7 +114,6 @@ void AC_AttitudeControl::relax_attitude_controllers()
     // TODO add _ahrs.get_quaternion()
     _attitude_target_quat.from_rotation_matrix(_ahrs.get_rotation_body_to_ned());
     _attitude_target_quat.to_euler(_attitude_target_euler_angle.x, _attitude_target_euler_angle.y, _attitude_target_euler_angle.z);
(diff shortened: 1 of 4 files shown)


libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
```diff
@@ -110,18 +110,26 @@ void AC_AttitudeControl::set_throttle_out_unstabilized(float throttle_in, bool r
 // Ensure attitude controller have zero errors to relax rate controller output
 void AC_AttitudeControl::relax_attitude_controllers()
 {
+    // Initialize the attitude variables to the current attitude
     // TODO add _ahrs.get_quaternion()
     _attitude_target_quat.from_rotation_matrix(_ahrs.get_rotation_body_to_ned());
```
Aksellence, line 115: The code seems to have no problem, but could you explain how this new logic should make a difference?

Aksellence (changes requested):

libraries/AC_AttitudeControl/AC_AttitudeControl.cpp
jschall: ```_attitude_target_euler_rate``` is not initialized in the original code.

Aksellence: Okay!

Aksellence (approved):

PR #44, opened 2019-02-22
Feature/gps switching logic fix

Aksellence:
Changed the Switching Logic between the two GPS's. Switches only when necessary. Tested in flight, and the frequency of switching was negligible.

### DIFF
--- a/libraries/AP_GPS/AP_GPS.cpp
+++ b/libraries/AP_GPS/AP_GPS.cpp
@@ -693,15 +693,27 @@ void AP_GPS::update(void)
                 // handle switch between real GPSs
                 for (uint8_t i=0; i<GPS_MAX_RECEIVERS; i++) {
                     if (i == primary_instance) {
-                        continue;
+                        continue; 
                     }
                     if (state[i].status > state[primary_instance].status) {
                         // we have a higher status lock, or primary is set to the blended GPS, change GPS
-                        primary_instance = i;
-                        _last_instance_swap_ms = now;
-                        continue;
+                        
+                        if (state[primary_instance].status < GPS_OK_FIX_3D) {
+                            primary_instance = i;
+                            _last_instance_swap_ms = now;
+                            continue;
+                        }
+                        // switch only if the currently used GPS has a 2D FIX or lower
+
+                        if (state[i].status > GPS_OK_FIX_3D) {
+                            primary_instance = i;
+                            _last_instance_swap_ms = now;
+                            continue;
+                        }
+                        // AND only if the other GPS has a DGPS FIX (the best possible fix)
                     }
 
+            /*
                     bool another_gps_has_1_or_more_sats = (state[i].num_sats >= state[primary_instance].num_sats + 1);
 
                     if (state[i].status == state[primary_instance].status && another_gps_has_1_or_more_sats) {
@@ -719,6 +731,8 @@ void AP_GPS::update(void)
                             _last_instance_swap_ms = now;
                         }
                     }
+            */
+
                 }
             }
         } else {
--- a/libraries/AP_GPS/AP_GPS.cpp
+++ b/libraries/AP_GPS/AP_GPS.cpp
@@ -701,6 +701,8 @@ void AP_GPS::update(void)
                         if (state[primary_instance].status < GPS_OK_FIX_3D) {
                             primary_instance = i;
                             _last_instance_swap_ms = now;
+                            gcs().send_text(MAV_SEVERITY_CRITICAL, "GPS Switch: Switched to %u", primary_instance+1);
+                            // Add alert when a switch happens
                             continue;
                         }
                         // switch only if the currently used GPS has a 2D FIX or lower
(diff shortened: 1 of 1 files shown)


libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -693,23 +693,35 @@ void AP_GPS::update(void)
                 // handle switch between real GPSs
                 for (uint8_t i=0; i<GPS_MAX_RECEIVERS; i++) {
                     if (i == primary_instance) {
-                        continue;
+                        continue;
```
presmini: remove space after semi-colon

libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -693,23 +693,35 @@ void AP_GPS::update(void)
...
+                            continue;
+                        }
+                        // switch only if the currently used GPS has a 2D FIX or lower
+
+                        if (state[i].status > GPS_OK_FIX_3D) {
+                            primary_instance = i;
+                            _last_instance_swap_ms = now;
+                            continue;
+                        }
+                        // AND only if the other GPS has a DGPS FIX (the best possible fix)
```
presmini: Is this logic currently really an AND logic? Looks to me as if the GPS will switch if the current primary is worse than 3D Fix OR the new primary is better than 3D fix. Doesn't this mean that you could end up switch to the new GPS if both had less than 3D fix?

presmini (changes requested):
There are a few questions I have here. I also think we should do a design review on the redundant GPS logic. Both the Mavnet side, and the Ardupilot side. Can you and @cvh10 put together documentation in Confluence, and we can have a design review?

libraries/AP_GPS/AP_GPS.cpp
Aksellence: No, it is not an 'AND' logic. In the comments, I wrote 'AND' just to showcase another scenario in which it will switch. I did not mean the 'AND' by the 'AND logic operator'.
So the logic now states that, if the FIX type of the other GPS is better than the FIX type of the primary GPS, it will look into the possibility of switching to the other GPS. And once that condition is met, there are only 2 possibilities in which the GPS switching will take place, i.e., 
1. It will switch only when the currently used GPS has a FIX type of 2D or less than that (which we absolutely want to avoid).
2. If the FIX type of the other GPS which is not being used that instant has an SBAS fix, which is the best fix possible and we should always be seeking for. 

And in case both the GPS have less than 3D Fix, it will only switch if the other GPS has a better FIX than the one we are using currently. (For Eg. if the one we are using has a 2D Fix and the other one has NO FIX, it will not switch on the other hand if the one we are using currently fails in flight for some reason (hence, it's fix type will turn into NO FIX or NO GPS) so even if the other GPS has a 2D FIX, it will switch. And that is exactly what we want.

libraries/AP_GPS/AP_GPS.cpp
Aksellence: That's where second pair of eyes comes in ;) 
Thanks

Aksellence:
> There are a few questions I have here. I also think we should do a design review on the redundant GPS logic. Both the Mavnet side, and the Ardupilot side. Can you and @cvh10 put together documentation in Confluence, and we can have a design review?

@presmini  Sure

libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -693,23 +693,35 @@ void AP_GPS::update(void)
...
                     }
 
+            /*
```
cvh10: Why was this logic removed?

libraries/AP_GPS/AP_GPS.cpp
cvh10: I don't see any timing considerations - it looks to me that it's possible that we could end up switching back and forth rapidly between the two receivers when they're both SBAS. I think that in the situation where both receivers have an SBAS fix type, there should be no switching or switching takes place only if the quality of the non-primary receiver is consistently better than the primary instance for an extended period of time.

libraries/AP_GPS/AP_GPS.cpp
Aksellence: The ideology behind the new logic is that there is no need to switch unless it is absolutely necessary to. Because the second GPS is installed as a back-up unit, to be used unless the primary GPS fails or gets really inaccurate.
So in order to ascertain that ideology, there seems to be no reason to use 'no. of satellites' as a factor for switching, rather, using just the 'FIX type' is more than enough to achieve our goal while maintaining flight safety and significantly lower no. of switching between the two GPS.

libraries/AP_GPS/AP_GPS.cpp
Aksellence: @cvh10  You are missing the main piece here. It won't even go into the switching logic unless the other GPS has a better Fix then the one used currently. So if the primary GPS has an SBAS fix, which is the maximum, it won't even go into the switching logic. 
So No, it won't keep switching between the 2 GPS's if both of them have an SBAS fix.
Also, ya I get your point on timing considerations, but that is a preference from a safety standpoint and the previous flight trend. I would prefer it to switch to an SBAS fix the moment the other GPS has that, rather for it to be waiting for 'x' seconds.

cvh10:
Looks good to me

PR #45, opened 2019-03-06
Fix to compass switching logic fix that can lead to increases error i…

Aksellence:
This fixes a critical bug in EKF compass switching that happens when you have exactly 3 healthy compasses (including internal and external compasses). The lack of the break in the loop resulted in the EKF always switching back to the compass that has been rejected due to high innovations immediately, which means compass failover due to innovations did not work.  
If you had 2 compasses then it worked correctly.


tridge (approved):
looks good, thanks Akshay!

PR #46, opened 2019-04-23
Added pre-arm and pre-takeoff checks for GPS quality

tridge:
This adds new Matternet specific parameters for checking GPS quality at both arming and at takeoff. The new parameters are:
 TOFF_POS_CHANGE: allowable position change between arming and takeoff
 TOFF_ALT_CHANGE: allowable altitude change between arming and takeoff
 ARM_GPS_HACC: max GPS reported horizontal accuracy for arming
 ARM_GPS_VACC: max GPS reported vertical accuracy for arming
The TOFF_* parameters apply when a MISSION_START is issued for automatic takeoff. The ARM_GPS_* parameters are applied at arming time.
In addition this PR includes changes to allow for SITL testing of GPS drift pre-takeoff. Using the SIM_GPS_DRIFT_S parameter the drift speed in m/s can be set. 

These pre-arm and pre-takeoff checks by themselves are not enough for full protection in poor GPS environments. In addition we need to prevent sudden attitude changes after takeoff is initiated due to changes in reported GPS position. That is achieved with a new WP_NAVALT_MAX parameter. That works like this:

 - below WP_NAVALT_MIN the attitude of the vehicle is kept completely level.
 - between WP_NAVALT_MIN and WP_NAVALT_MAX the allowable attitude control is linearly increases up to ANGLE_MAX
 - above WP_NAVALT_MAX the full ANGLE_MAX is available

this change prevents a sudden change in allowable attitude for position control which greatly reduces the chance of attitude overshoot at low altitude. The selection of WP_NAVALT_MIN and WP_NAVALT_MAX is highly dependent on the maximum wind conditions that the vehicle will operate in, and the proximity to obstacles that provide either a physical risk or a risk of GPS interference

This PR also includes a change to fix handling of an IMU failure. This is a critical fix for any vehicle using a Hex Cube.

### DIFF
--- a/ArduCopter/AP_Arming.cpp
+++ b/ArduCopter/AP_Arming.cpp
@@ -428,7 +428,7 @@ bool AP_Arming_Copter::gps_checks(bool display_failure)
 
     float hacc, vacc;
     if (!copter.gps.horizontal_accuracy(hacc) ||
-        !copter.gps.horizontal_accuracy(vacc)) {
+        !copter.gps.vertical_accuracy(vacc)) {
         if (display_failure) {
             gcs().send_text(MAV_SEVERITY_CRITICAL,"PreArm: No GPS accuracy");
         }
--- a/ArduCopter/AP_Arming.cpp
+++ b/ArduCopter/AP_Arming.cpp
@@ -426,6 +426,27 @@ bool AP_Arming_Copter::gps_checks(bool display_failure)
         return false;
     }
 
+    float hacc, vacc;
+    if (!copter.gps.horizontal_accuracy(hacc) ||
+        !copter.gps.horizontal_accuracy(vacc)) {
+        if (display_failure) {
+            gcs().send_text(MAV_SEVERITY_CRITICAL,"PreArm: No GPS accuracy");
+        }
+        return false;
+    }
+    if (hacc > copter.matternet.arm_gps_hacc) {
+        if (display_failure) {
+            gcs().send_text(MAV_SEVERITY_CRITICAL, "PreArm: GPS hacc %.2f", hacc);
+        }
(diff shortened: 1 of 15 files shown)


ArduCopter/AP_Arming.cpp
```diff
@@ -426,6 +426,27 @@ bool AP_Arming_Copter::gps_checks(bool display_failure)
         return false;
     }
 
+    float hacc, vacc;
+    if (!copter.gps.horizontal_accuracy(hacc) ||
+        !copter.gps.horizontal_accuracy(vacc)) {
```
Aksellence: Shouldn't it be " !copter.gps.**vertical_accuracy**(vacc)) { " ?

Aksellence (changes requested):
Looks good except for that one line. Seem like a small bug.

presmini (changes requested):
overall looks good except change Akshay caught. I'm curious how we are going to determine what are the acceptable levels for each of these parameters. Any thoughts? Is there a per-location site survey procedure we need to implement to ensure we have the "correct" values for these new thresholds?

ArduCopter/AP_Arming.cpp
tridge: yes, you're right, well spotted! I'll fix now

tridge: fixed

tridge:
@presmini I have suggested a 24hr site survey, but I don't know how practical that is with the matternet operations. The acceptable levels also depend on maximum wind we will operate in, and how close obstacles are. If obstacles are far away then we can have large values for the limits, but we also are less likely to hit those limits.

PR #47, opened 2019-04-25
Clamp rate integrators below WP_NAVALT_MIN

tridge:
This prevents a small AHRS_TRIM error causing rate integrator buildup during the first phase of takeoff where we don't have good attitude control.
In testing that integrator buildup can cause a 5 degree pitch overshoot when the landing gear does leave the ground


Aksellence (approved):
Looks good!

presmini (approved):
LGTM

PR #50, opened 2019-05-07
Copter: disable RC arming check in GUIDED mode

tridge:
This allows for auto missions with RC off, while enabling RC arming checks in manual modes
The vehicle should be setup with:
 ARMING_CHECK=1
 FS_THR_ENABLE=6
the FS_THR_ENABLE=6 is to give the behaviour that RC fail while in AUTO is ignored, but in manual modes it does LAND
This PR also fixes the issue with needing to arm twice in GUIDED mode if RC is off


Aksellence (approved):
Ya.. Looks good to me.

PR #52, opened 2019-05-09
Copter: don't check GPS quality on MISSION_START if flying

tridge:
MISSION_START can be used to move to a changed mission for landing. In
that case the GPS pre-takeoff checks are not appropriate


Aksellence (approved):
Looks good.

PR #53, opened 2019-05-17
Force first EKF lane and report worst variance over MAVLink

tridge:
This forces the first EKF lane while disarmed, preventing a condition where a lane switch can happen while waiting for good GPS lock.
It also changes the MAVLink EKF_STATUS_REPORT message to show the maximum of all scaled innovations across all lanes, so that an existing MAVLink GCS can know if one of the lanes is unhealthy

### DIFF
--- a/libraries/AP_NavEKF2/AP_NavEKF2.cpp
+++ b/libraries/AP_NavEKF2/AP_NavEKF2.cpp
@@ -1252,11 +1252,70 @@ void  NavEKF2::getFilterGpsStatus(int8_t instance, nav_gps_status &status)
 }
 
 // send an EKF_STATUS_REPORT message to GCS
+// for each variance we send the maximum of all lanes
 void NavEKF2::send_status_report(mavlink_channel_t chan)
 {
-    if (core) {
-        core[primary].send_status_report(chan);
+    nav_filter_status filterStatus = core[primary].filterStatus;
+
+    // prepare flags
+    uint16_t flags = 0;
+    if (filterStatus.flags.attitude) {
+        flags |= EKF_ATTITUDE;
     }
+    if (filterStatus.flags.horiz_vel) {
+        flags |= EKF_VELOCITY_HORIZ;
+    }
+    if (filterStatus.flags.vert_vel) {
+        flags |= EKF_VELOCITY_VERT;
+    }
+    if (filterStatus.flags.horiz_pos_rel) {
+        flags |= EKF_POS_HORIZ_REL;
+    }
+    if (filterStatus.flags.horiz_pos_abs) {
+        flags |= EKF_POS_HORIZ_ABS;
+    }
+    if (filterStatus.flags.vert_pos) {
+        flags |= EKF_POS_VERT_ABS;
+    }
+    if (filterStatus.flags.terrain_alt) {
+        flags |= EKF_POS_VERT_AGL;
+    }
+    if (filterStatus.flags.const_pos_mode) {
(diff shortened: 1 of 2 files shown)


libraries/AP_NavEKF2/AP_NavEKF2.cpp
```diff
@@ -765,6 +765,12 @@ void NavEKF2::UpdateFilter(void)
         }
     }
 
+    // when disarmed force lane1 if it is healthy. This is a temporary
+    // workaround for poor performance of the 2nd IMU
+    if (!hal.util->get_soft_armed() && primary != 0 && core[0].healthy()) {
```
Aksellence: Looks good. But seems like it can still switch to the other EKF Lane in case the lane1 is not healthy, rather, not using any lane at all. Is this what we want, or should we deny switching to any lane completely in case the 1st lane is not ready when the vehicle is in soft_armed mode?

libraries/AP_NavEKF2/AP_NavEKF2.cpp
```diff
@@ -1252,11 +1252,70 @@ void  NavEKF2::getFilterGpsStatus(int8_t instance, nav_gps_status &status)
...
+        velVar = MAX(velVar, c_velVar);
+        hgtVar = MAX(hgtVar, c_hgtVar);
+        tasVar = MAX(tasVar, c_tasVar);
+        magVarLen = MAX(magVarLen, c_magVar.length());
+    }
+
+    // Only report range finder normalised innovation levels if the EKF needs the data for primary
+    // height estimation or optical flow operation. This prevents false alarms at the GCS if a
+    // range finder is fitted for other applications
+    float temp;
```
Aksellence, line 1310: I think we should rename it to something more specific?

Aksellence (commented):
Everything looks good except of a few minor things.

libraries/AP_NavEKF2/AP_NavEKF2.cpp
tridge: It is a copy of existing code from NavEKF2_core::send_status_report. I wouldn't have chosen that var name, but I want to keep the code as similar as possible

Aksellence: Okay

libraries/AP_NavEKF2/AP_NavEKF2.cpp
Aksellence: What about this?

libraries/AP_NavEKF2/AP_NavEKF2.cpp
```diff
@@ -767,7 +767,7 @@ void NavEKF2::UpdateFilter(void)
 
     // when disarmed force lane1 if it is healthy. This is a temporary
     // workaround for poor performance of the 2nd IMU
-    if (!hal.util->get_soft_armed() && primary != 0 && core[0].healthy()) {
+    if (!hal.util->get_soft_armed() && primary != 0) {
```
Aksellence, line 770: Just want to make sure, now that it will always stay in 1st lane, even when it is unhealthy. The vehicle will not be able to pass the pre-arm checks in case lane 1 is unhealthy, right?
Because I don't want the vehicle to be capable of arming even when the EKF lane it is on is unhealthy.

Aksellence (commented):
Looks good!

Aksellence (approved):

PR #54, opened 2019-06-07
Fixed yaw in NAV_LAND

tridge:
This fixes yaw handling in NAV_LAND in AUTO mode. The fixed yaw from a CONDITION_YAW was being ignored

### DIFF
--- a/ArduCopter/mode_auto.cpp
+++ b/ArduCopter/mode_auto.cpp
@@ -876,7 +876,7 @@ void Copter::ModeAuto::land_run()
     // set motors to full range
     motors->set_desired_spool_state(AP_Motors::DESIRED_THROTTLE_UNLIMITED);
     
-    land_run_horizontal_control();
+    land_run_horizontal_control(auto_yaw.mode()==AUTO_YAW_FIXED, auto_yaw.yaw());
     land_run_vertical_control();
 }
 
--- a/Tools/scripts/runcoptertest.py
+++ b/Tools/scripts/runcoptertest.py
@@ -35,7 +35,7 @@ def wait_time(mav, simtime):
         if t2 - t1 > simtime:
             break
 
-cmd = '../Tools/autotest/sim_vehicle.py -j4 -D'
+cmd = '../../Tools/autotest/sim_vehicle.py -D -S10'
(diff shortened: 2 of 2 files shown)


ArduCopter/mode_auto.cpp
```diff
@@ -876,7 +876,7 @@ void Copter::ModeAuto::land_run()
     // set motors to full range
     motors->set_desired_spool_state(AP_Motors::DESIRED_THROTTLE_UNLIMITED);
     
-    land_run_horizontal_control();
+    land_run_horizontal_control(auto_yaw.mode()==AUTO_YAW_FIXED, auto_yaw.yaw());
```
Aksellence, line 879: Ah.. I see why we using this. But still my question lies, how were we able to compile the method without the params in the first place, as on looking at the repo I couldn't see any polymorphic function there (without the params)?

Tools/scripts/runcoptertest.py
```diff
@@ -35,24 +35,25 @@ def wait_time(mav, simtime):
...
+mavproxy.send('param set SIM_SPEEDUP 5\n')
 wait_mode(mav, ['GUIDED'])
 mavproxy.expect('is using GPS')
 mavproxy.expect('is using GPS')
 mavproxy.send('arm throttle\n')
 mavproxy.expect('ARMED')
 mavproxy.send('mode AUTO\n')
 wait_mode(mav, ['AUTO'])
-wait_time(mav, 30)
+wait_time(mav, 15)
```
Aksellence, line 54: Why did we reduce this time?

ArduCopter/mode_auto.cpp
tridge: it takes optional parameters (ie parameters with default values)

Tools/scripts/runcoptertest.py
tridge: this is just my little test script, 15s is closer to what we normally have between arming and takeoff. 30s was for when I was testing gps drift

Aksellence: Okay Sounds good!

Aksellence (approved):

PR #57, opened 2019-06-07
Fixed mismatch of EKF lane switching logic

tridge:
This ensures that if we are about to declare an EKF failsafe (and trigger parachute) that we first check to see if we can instead change EKF lanes to avoid the error

### DIFF
--- a/libraries/AP_NavEKF2/AP_NavEKF2.cpp
+++ b/libraries/AP_NavEKF2/AP_NavEKF2.cpp
@@ -740,10 +740,52 @@ void NavEKF2::UpdateFilter(void)
             updateLaneSwitchPosResetData(newPrimaryIndex, primary);
             updateLaneSwitchPosDownResetData(newPrimaryIndex, primary);
             primary = newPrimaryIndex;
+            lastLaneSwitch_ms = AP_HAL::millis();
         }
     }
 
     check_log_write();
+}
+
+/*
+  check if switching lanes will reduce the normalised
+  innovations. This is called when the vehicle code is about to
+  trigger an EKF failsafe, and it would like to avoid that by
+  using a different EKF lane
+*/
+void NavEKF2::checkLaneSwitch(void)
+{
+    uint32_t now = AP_HAL::millis();
+    if (lastLaneSwitch_ms != 0 && now - lastLaneSwitch_ms < 5000) {
+        // don't switch twice in 5 seconds
+        return;
+    }
+    float primaryErrorScore = core[primary].errorScore();
+    float lowestErrorScore = primaryErrorScore;
+    uint8_t newPrimaryIndex = primary;
+    for (uint8_t coreIndex=0; coreIndex<num_cores; coreIndex++) {
+        if (coreIndex != primary) {
+            // an alternative core is available for selection only if healthy and if states have been updated on this time step
+            bool altCoreAvailable = core[coreIndex].healthy();
+            float altErrorScore = core[coreIndex].errorScore();
+            if (altCoreAvailable &&
+                altErrorScore < lowestErrorScore &&
+                altErrorScore < 0.9) {
+                newPrimaryIndex = coreIndex;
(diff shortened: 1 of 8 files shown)


libraries/AP_NavEKF2/AP_NavEKF2.cpp
```diff
@@ -740,12 +740,54 @@ void NavEKF2::UpdateFilter(void)
...
+    float lowestErrorScore = primaryErrorScore;
+    uint8_t newPrimaryIndex = primary;
+    for (uint8_t coreIndex=0; coreIndex<num_cores; coreIndex++) {
+        if (coreIndex != primary) {
+            // an alternative core is available for selection only if healthy and if states have been updated on this time step
+            bool altCoreAvailable = core[coreIndex].healthy();
+            float altErrorScore = core[coreIndex].errorScore();
+            if (altCoreAvailable &&
+                altErrorScore < lowestErrorScore &&
+                altErrorScore < 0.9) {
```
Aksellence, line 773: The threshold to deploy a parachute due to EKF variance is 0.8. Shouldn't this errorScore be reduced to <0.7 or <0.75?

libraries/AP_NavEKF3/AP_NavEKF3.cpp
```diff
@@ -778,12 +778,52 @@ void NavEKF3::UpdateFilter(void)
...
+    }
+    float primaryErrorScore = core[primary].errorScore();
+    float lowestErrorScore = primaryErrorScore;
+    uint8_t newPrimaryIndex = primary;
+    for (uint8_t coreIndex=0; coreIndex<num_cores; coreIndex++) {
+        if (coreIndex != primary) {
+            // an alternative core is available for selection only if healthy and if states have been updated on this time step
+            bool altCoreAvailable = core[coreIndex].healthy();
+            float altErrorScore = core[coreIndex].errorScore();
+            if (altCoreAvailable && altErrorScore < lowestErrorScore && altErrorScore < 0.9) {
```
Aksellence, line 809: The threshold to deploy a parachute due to EKF variance is 0.8. Shouldn't this errorScore be reduced to <0.7 or <0.75?

Aksellence (changes requested):

libraries/AP_NavEKF2/AP_NavEKF2.cpp
tridge: no, the two error scores don't mean the same thing and aren't calculated in the same way. What this is checking is that the lane we're switching to is significantly better than the lane we're on.

libraries/AP_NavEKF3/AP_NavEKF3.cpp
tridge: see above

libraries/AP_NavEKF2/AP_NavEKF2.cpp
Aksellence: Got it. Then it's fine

Aksellence (approved):

libraries/AP_NavEKF2/AP_NavEKF2.cpp
```diff
@@ -740,12 +740,54 @@ void NavEKF2::UpdateFilter(void)
...
+    }
+
+    // update the yaw and position reset data to capture changes due to the lane switch
+    if (newPrimaryIndex != primary) {
+        updateLaneSwitchYawResetData(newPrimaryIndex, primary);
+        updateLaneSwitchPosResetData(newPrimaryIndex, primary);
+        updateLaneSwitchPosDownResetData(newPrimaryIndex, primary);
+        primary = newPrimaryIndex;
+        lastLaneSwitch_ms = now;
+        gcs().send_text(MAV_SEVERITY_CRITICAL, "NavEKF2: lane switch %u", primary);
```
Aksellence, line 787: Should we change this message to something which signals that the vehicle was about to FTS? Rather than making it a bit generic?

PR #58, opened 2019-06-10
Persist logging after disarm or arming failure

tridge:
This persists logging for 15s after logging failure or disarm. This will give more information on the reasons for an arming failure and more information on an unexpected disarm

### DIFF
--- a/libraries/DataFlash/DataFlash.cpp
+++ b/libraries/DataFlash/DataFlash.cpp
@@ -23,6 +23,13 @@ extern const AP_HAL::HAL& hal;
 #define HAL_DATAFLASH_MAV_BUFSIZE  8
 #endif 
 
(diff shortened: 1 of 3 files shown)


libraries/DataFlash/DataFlash.cpp
```diff
@@ -23,6 +23,13 @@ extern const AP_HAL::HAL& hal;
 #define HAL_DATAFLASH_MAV_BUFSIZE  8
 #endif 
 
+// by default log for 15 seconds after disarming
```
Aksellence: 15 seconds to 60 seconds

Aksellence (changes requested):
Looks good except for a tiny comment change.

Aksellence (approved):

Aksellence (approved):

PR #62, opened 2019-07-02
Toos: added bootloader for MttrCubeBlack

tridge:
this is included in ROMFS, for updating bootloader with mavlink.
Note that the board needs about 22k free memory to flash the bootloader


Aksellence (commented):
Can't see anything here.

PR #63, opened 2019-07-03
GCS_MAVLink: break out MTTR_FTS_VER string send

tridge:
and ensure it is sent on both capabilities and version request


Aksellence (approved):
Looks fine
Tested and works fine.

Aksellence (approved):

PR #65, opened 2019-07-04
AP_Notify: play error tone if we have no microSD at boot

tridge:


Aksellence:
@tridge Why does it fail all the travis-ci builds?

Aksellence (approved):

PR #68, opened 2019-07-19
Pr set f9p primary gps

Aksellence:
Added logic for choosing F9P over other GPS as a primary GPS.

### DIFF
--- a/libraries/AP_GPS/AP_GPS.cpp
+++ b/libraries/AP_GPS/AP_GPS.cpp
@@ -739,26 +739,34 @@ void AP_GPS::update(void)
                     if (i == primary_instance) {
                         continue;
                     }
-                    if (state[i].status > state[primary_instance].status) {
-                        // we have a higher status lock, or primary is set to the blended GPS, change GPS
-                        
-                        if (state[primary_instance].status < GPS_OK_FIX_3D) {
-                            primary_instance = i;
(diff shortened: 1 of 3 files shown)


libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -739,26 +739,34 @@ void AP_GPS::update(void)
...
+                    }
+
+                    // Treat U-blox F9 as SBAS as it is dual band GPS and is more accurate than M8 with SBAS                    
+                    if (status_primary == GPS_OK_FIX_3D && strcmp(drivers[primary_instance]->name(), "u-blox") == 0 && drivers[primary_instance]->hardware_generation() == AP_GPS_UBLOX::UBLOX_F9) {
+                        status_primary = GPS_OK_FIX_3D_DGPS;
+                    }
+
+                    bool should_switch = false;
+
+                    if (status_i > status_primary) {
```
Aksellence, line 758: @tridge I think we should change this part of the logic, because it seems like the vehicle can choose a bad F9P in case the M8 has a 3D fix. I am imagining a scenario where, the F9P is seeing only 6-9 satellites, and on the other hand the M8 is seeing 14 satelltes (but has a 3D fix).
As we hard code the F9P to be considered as SBAS always, we would still choose F9P over M8, based on the fix type and not taking into consideration the number of satellites at all.

PR #76, opened 2019-08-07
Tools: update MttrCubeBlack bootloader to listen on only telem1

tridge:
this disables bootloading from telem2 and GPS ports


Aksellence (approved):
Cannot view the file. But saw you making changed in front of me when was on a screen shared call, so approved.

PR #77, opened 2019-08-14
AP_NavEKF2: added EK2_MAG_LRN_LIM

tridge:
This sets a limit on EKF2 magnetometer offset learning. This is an attempt to cope with sites where localized magnetic fields are causing significant innovations which trigger learned biases which take several minutes to recover. This PR is currently for testing purposes only, it is possible that limiting the offsets in this way will lead to the errors being pushed into other states of the EKF which could lead to instability. We are already constraining the earth field states against the WMM, so the error may be pushed into bias states for other sensors, or just produce an incorrect yaw estimate.
I suggest we start with EK2_MAG_LRN_LIM=50 and see if that produces a sufficient reduction in the incorrect mag bias learning. 
This change does rely on having well calibrated compasses. It will reduce our ability to cope with poor initial compass offsets


Aksellence (approved):
Looks good.

PR #81, opened 2019-08-15
3.6.x fixes from 3.6.10 and pending critical PRs

tridge:
This brings in some critical fixes from the 3.6.x development branch, plus fixes from the IOMCU PR that is pending against master
Changes are:
 - fixed handling of unhealthy baro on CubeBlack
 - added learning of inactive gyro biases
 - fixed lost IMU samples under load
 - fixed IOMCU UART corruption
 - update to latest IO firmware


libraries/AP_HAL_ChibiOS/UARTDriver.cpp
```diff
@@ -530,8 +530,12 @@ size_t UARTDriver::write(const uint8_t *buffer, size_t size)
 		return 0;
 	}
 
-    if (!_write_mutex.take_nonblocking()) {
-        return 0;
+    if (_blocking_writes && unbuffered_writes) {
+        _write_mutex.take_blocking();
+    } else {
```
Aksellence, line 535: Can we do this an 'else if' ?

Aksellence (approved):

libraries/AP_HAL_ChibiOS/UARTDriver.cpp
tridge: why? what would the condition be on?

tridge: yes, we could, but I'd like this to match upstream

PR #89, opened 2019-09-26
Added table based SoC system when disarmed

tridge:
This adds a new BATT_MONITOR option 20, which selects a table based SoC system for battery monitoring when disarmed. It uses a table that is specific to Matternet vehicles


Aksellence (approved):
Looks good.

PR #90, opened 2019-09-26
Fix GPS outage bug

tridge:
This fixes a bug in u-blox protocol parsing that could cause a single lost byte to lead to GPS disconnect and reset. The problem happens if the bytes after the lost bytes contain a pattern that is formed like a u-blox packet with zero payload size. In that case we would process several kByte of data before finally resetting the device 4 seconds later.
This also adds logging of the event in a new GPST message, along with logging of CRC errors, resets and total traffic on GPS UARTs


Aksellence (approved):
looks good.

PR #91, opened 2019-10-02
Fixed logging with auto log rotation on disarm

tridge:
This fixes an issue APM-34 where we would put data from the 60s post disarm logging in a new log file if LOG_FILE_DSRMROT=1.
It also triggers a 2nd write of the system IDs and parameters on disarm, ensuring that even with log trimming we always get the system IDs (git version, board ID etc)


Aksellence (approved):
Yea. Looks good.

PR #93, opened 2019-10-05
Added logging of BMU battery ID

tridge:
This will let us associate a battery serial number with each flight in the ArduPilot log


libraries/AP_UAVCAN/AP_UAVCAN.h
```diff
@@ -118,6 +118,8 @@ class AP_UAVCAN {
         float remaining_capacity_wh;
         float full_charge_capacity_wh;
         uint8_t status_flags;
+        uint32_t model_instance_id;
+        uint8_t model_name[32+1];
```
Aksellence, line 122: Can we have more insight into what does '32' and '1' stand for? We could define them as a const maybe?

libraries/AP_UAVCAN/AP_UAVCAN.cpp
```diff
@@ -1338,7 +1338,7 @@ uint8_t AP_UAVCAN::register_BM_bi_listener_to_id(AP_BattMonitor_Backend* new_lis
     }
 
     for (uint8_t i = 0; i < AP_UAVCAN_MAX_BI_NUMBER; i++) {
-        if (_bi_id[i] != id) {
+        if (id != -1 && _bi_id[i] != id) {
```
Aksellence, line 1341: It would be good to define -1 as a const

Aksellence (approved):
Other than the few comments on the const, everything looks good.

PR #94, opened 2019-10-21
Added support for new MttrCubeBlack+ board

tridge:
Note that this needs to be a new build as it has a different board ID in the bootloader


Aksellence (approved):
Looks good.

zalice:
Moving from cube black to cube orange. Not using black+

PR #95, opened 2019-10-22
Add support for RTCM injection to UAVCAN GPS modules

tridge:
This allows for NTRIP RTCM data to be used with UAVCAN GPS modules


Aksellence (approved):
Looks good.

PR #96, opened 2019-10-24
AP_GPS: fixed GPS to UART mapping

tridge:
this allows for first GPS as UAVCAN and 2nd as UART. That is not
possible currently unless you waste a uart


Aksellence (approved):
Looks good. The next steps for the future should be to make the GPS switching logic more robust.

PR #97, opened 2019-10-25
Copter: improve land_detector rangefinder check

tridge:
we consider the check met if any healthy rangefinder is below the min
alt threshold. This fixes an issue with radars which can give crazy
data close to the ground


Aksellence (approved):
Looks good.

PR #99, opened 2019-11-21
Added support for fixed yaw mag calibration

tridge:
This is a fast compass calibration given vehicle position and yaw. The calibrations results in zero diagonal and off-diagonal elements, so is only suitable for vehicles where the field is close to spherical. It is useful for large vehicles where moving the vehicle to calibrate it is difficult.

The offsets of the selected compasses are set to values to bring them into consistency with the WMM tables at the given latitude and longitude. If compass_mask is zero then all enabled compasses are calibrated.

This assumes that the compass is correctly scaled in milliGauss

MAVProxy implements this using the "magcal yaw" command.

This PR depends on this change to mavlink:
https://github.com/matternet/mavlink/pull/3

UPDATE
This PR now also includes the new COMPASS_SCALE estimation, allowing us to compensate for incorrectly scaled compasses. On a full 3D compass cal the COMPASS_SCALE will be automatically estimated. If COMPASS_SCALE is not set (ie. it is zero) then the use of the WMM constraints in the EKF are disabled

### DIFF
--- a/libraries/AP_Compass/CompassCalibrator.cpp
+++ b/libraries/AP_Compass/CompassCalibrator.cpp
@@ -461,11 +461,6 @@ float CompassCalibrator::calc_residual(const Vector3f& sample, const param_t& pa
     return params.radius - (softiron*(sample+params.offset)).length();
 }
 
-float CompassCalibrator::calc_mean_squared_residuals() const
-{
-    return calc_mean_squared_residuals(_params);
-}
-
 // calc the fitness given a set of parameters (offsets, diagonals, off diagonals)
 float CompassCalibrator::calc_mean_squared_residuals(const param_t& params) const
 {
--- a/libraries/AP_Compass/CompassCalibrator.cpp
+++ b/libraries/AP_Compass/CompassCalibrator.cpp
@@ -73,10 +73,10 @@ CompassCalibrator::CompassCalibrator():
     _tolerance(COMPASS_CAL_DEFAULT_TOLERANCE),
     _sample_buffer(nullptr)
 {
-    clear();
+    stop();
 }
 
-void CompassCalibrator::clear()
+void CompassCalibrator::stop()
 {
     set_status(COMPASS_CAL_NOT_STARTED);
 }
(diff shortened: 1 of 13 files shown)


libraries/AP_Compass/CompassCalibrator.cpp
```diff
@@ -467,6 +482,17 @@ float CompassCalibrator::calc_mean_squared_residuals(const param_t& params) cons
     return sum;
 }
 
+// calculate initial offsets by simply taking the average values of the samples
+void CompassCalibrator::calc_initial_offset()
+{
+    // Set initial offset to the average value of the samples
+    _params.offset.zero();
+    for (uint16_t k = 0; k < _samples_collected; k++) {
+        _params.offset -= _sample_buffer[k].get();
```
Aksellence, line 499: Shouldn't this be '+=' rather than '-=' ? Or this actually needs to be a difference between the Offset and the sample offset val?

libraries/AP_Compass/Compass_learn.cpp
```diff
@@ -108,10 +116,27 @@ void CompassLearn::update(void)
         if (num_samples > 30 && best_error < 50 && worst_error > 65) {
```
Aksellence, line 115: shouldn't we have a limit here for the 'worst_error'? It seems like any value above 65 is acceptable

libraries/AP_Compass/CompassCalibrator.cpp
tridge: it needs to be -= as we are finding the value which when added to the samples gives zero

libraries/AP_Compass/Compass_learn.cpp
tridge: we're not using this in-flight learning code in MatterNet

Aksellence: Alright.

libraries/AP_Compass/CompassCalibrator.cpp
Aksellence: Gotcha, thanks!

Aksellence (approved):

PR #104, opened 2019-12-18
Fixed I2C interrupt handling in ChibiOS

tridge:
This fixes a recently discovered issue with I2C interrupt handling in ChibiOS on STM32F4xx MCUs. The fix includes both handling the required interrupt combinations and adding an interrupt quota per byte to prevent any future similar issues from causing a lockup


Aksellence (approved):
Looks good.

PR #107, opened 2020-03-04
Added MIS_OPTIONS bit for continue after land

tridge:
This changes the default behaviour of mission process on the completion of a NAV_LAND. The current behaviour is to continue with the next mission item. That creates a problem for missions where there are multiple contingencies planned in one mission where each has it's own NAV_LAND.
With this PR the behaviour is changed so that on completion of a NAV_LAND the copter will disarm and stop mission processing. If MIS_OPTIONS=4 is set then the old behaviour is restored and the mission continues on landing completion.
This also changes the messages when landing. The code previously printed "Landed" continuously while descending. It now prints "Landed" only once when land is complete and prints "Landing" once when landing descent is started.

### DIFF
--- a/ArduCopter/mode_auto.cpp
+++ b/ArduCopter/mode_auto.cpp
@@ -1565,15 +1565,29 @@ bool Copter::ModeAuto::verify_land()
                 land_start(dest);
 
                 // advance to next state
+                if (land_state != LandStateType_Descending) {
+                    gcs().send_text(MAV_SEVERITY_INFO,"Landing");
+                }
                 land_state = LandStateType_Descending;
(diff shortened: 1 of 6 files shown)


zalice (approved):
Thank you, Tridge. Tested in SITL with our software stack and behaves as desired.

Aksellence:
Looks good.

Aksellence (approved):
Looks good.

Aksellence:
@tridge I am able to build it for 'CubeBlack' build but not build for 'MttrCubeBlack' build.

tridge:
@AkshayMatternet what error do you get with MttrCubeBlack?

Aksellence (approved):

PR #110, opened 2020-05-02
Added terrain generation script

tridge:
This allows for generation of terrain data which allows the GCS terrain protocol to be avoided

### DIFF
--- a/libraries/AP_Terrain/AP_Terrain.cpp
+++ b/libraries/AP_Terrain/AP_Terrain.cpp
@@ -47,6 +47,13 @@ const AP_Param::GroupInfo AP_Terrain::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("SPACING",   1, AP_Terrain, grid_spacing, 100),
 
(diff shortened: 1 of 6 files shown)


Aksellence (approved):
I can't see any noticeable issues. But would want to understand what exactly is the create_terrain script doing, and is going to work for our purposes.
Looks good.

tridge:
> But would want to understand what exactly is the create_terrain script doing, and is going to work for our purposes.

what do you need to know beyond what you can get from the help and reading the script?

zalice:
@AkshayMatternet does this make more sense after our discussion or do you have open questions?

Aksellence:
> @AkshayMatternet does this make more sense after our discussion or do you have open questions?

Yeah, all the questions I had were answered during the discussion. Thanks!

Aksellence (approved):

zalice (approved):

adriennet823 (approved):
Awesome! Ran this myself and it worked great.

PR #113, opened 2020-06-01
Update handle mission item upper bound

adriennet823:
Handle mission count updated from `packet.count-1` to `packet.count` as the former was resulting in a request for mission waypoints index start -> end-1 (full mission not sent). This change mirrors the waypoint indices that are loaded in with mission count in 3.6.7: https://github.com/matternet/ardupilot/blob/b8353e7ef57bbe3664cf03afbc3df7ca46b2c952/libraries/GCS_MAVLink/GCS_Common.cpp#L548
Additionally, handle mission item is range is updated to be inclusive of the upper `request_last` bound index, as was the case with in 3.6.7 https://github.com/matternet/ardupilot/blob/b8353e7ef57bbe3664cf03afbc3df7ca46b2c952/libraries/GCS_MAVLink/GCS_Common.cpp#L765 so that the same mission waypoints during a partial waypoint send are requested by AP from the GCS as with 3.6.7


Aksellence (approved):
Looks good to me.

rabad-matternet (approved):
:+1:

tridge (approved):
this looks fine and I think existing GCS implementations will handle it fine. I tested it with both MissionPlanner and mavproxy with no issues

adriennet823:
> this looks fine and I think existing GCS implementations will handle it fine. I tested it with both MissionPlanner and mavproxy with no issues

Awesome, thank you for testing!

PR #115, opened 2020-06-08
Reset baro calibration at 1Hz when on ground (for 3.6.7)

tridge:
As discussed with Alice

### DIFF
--- a/ArduCopter/ArduCopter.cpp
+++ b/ArduCopter/ArduCopter.cpp
@@ -436,6 +436,12 @@ void Copter::one_hz_loop()
         // set all throttle channel settings
         motors->set_throttle_range(channel_throttle->get_radio_min(), channel_throttle->get_radio_max());
 #endif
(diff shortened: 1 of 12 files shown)


ArduCopter/ArduCopter.cpp
```diff
@@ -436,6 +436,12 @@ void Copter::one_hz_loop()
         // set all throttle channel settings
         motors->set_throttle_range(channel_throttle->get_radio_min(), channel_throttle->get_radio_max());
 #endif
+
+        if (gps.status() >= AP_GPS::GPS_OK_FIX_3D) {
+            // reset baro cal when on the ground
+            barometer.update_calibration();
```
zalice, line 442: @tridge why is the baro cal reliant on a GPS fix?

zalice (approved):

adriennet823 (approved):

PR #116, opened 2020-06-24
SITL: tweaks to bring simulated vehicle closer to matternet vehicle

tridge:
This brings the current draw, voltage and hover throttle approximately in line with the real vehicle


rabad-matternet (approved):

adriennet823:
Tested this and observed a significant improvement (increase in power and speed) in SITL performance (in comparison to a build with `mttr-rebase-4.0`). @tridge if there are no other small tweaks you think we should make (including AP parameters), I'd like to get this merged in

adriennet823 (approved):

Aksellence (approved):
Looks good to me.

PR #117, opened 2020-06-26
AP_RangeFinder: cope with beyond max range with LightwareI2C

tridge:
This fixes parsing of "-1.00" as a return value for the Lightware I2C lidar
Also fixed 65436 from legacy parsing


Aksellence (approved):
Had a bug earlier in the legacy parser, fixed it, tested, everything looks good.

PR #119, opened 2020-07-21
AP_RangeFinder: fixed rangefinder status for UAVCAN

tridge:
this fixes use of the MIN_CM parameter for UAVCAN rangefinders


Aksellence (approved):
Tested and performs as expected. Looks good. Thanks!

PR #120, opened 2020-07-25
Work around magZ offset issues (for 3.6.7)

tridge:
This implements a feature in EKF2 that works around a compass Z offset issue seen on the vehicles. It works as follows:

 - the maximum learned Z mag offset given by EK2_MAG_LRN_LIM is made 3x as large on the Z axis, which aligned with the observed higher uncertainty in Z mag offsets in the vehicles
 - when the yaw alignment is done 2.5m above the ground on takeoff we set the initial learned Z mag offset to be the value given by the WMM table for the location. This corrects the Z axis of the compass for the changing Z field.

The background to this PR is a problem with the Z mag readings on the compass varying between flights. The Z mag offset has very low observability when the copter is level, but it also has a correspondingly low impact on the yaw estimate. Allowing the Z mag offset to vary by a higher amount and initialising it to the correct value will lower innovations and should result in better EKF accuracy


libraries/AP_NavEKF2/AP_NavEKF2_MagFusion.cpp
```diff
@@ -390,6 +391,17 @@ void NavEKF2_core::FuseMagnetometer()
         MagPred[1] = DCM[1][0]*magN + DCM[1][1]*magE  + DCM[1][2]*magD + magYbias;
         MagPred[2] = DCM[2][0]*magN + DCM[2][1]*magE  + DCM[2][2]*magD + magZbias;
 
+        if (pending_z_mag_correction && frontend->_mag_learn_limit > 0) {
+            pending_z_mag_correction = false;
+            float correction = magDataDelayed.mag[2] - MagPred[2];
```
rabad-matternet, line 396: I can see that *.mag and MagPred are defined as 3-element vectors.  Do we assume that indices 0, 1, and 2 are X, Y, and Z, respectively?  Should these be #defined, const, or constexpr so that it’s more clear??  E.g. MagPred[Z_AXIS]

libraries/AP_NavEKF2/AP_NavEKF2_MagFusion.cpp
```diff
@@ -390,6 +391,17 @@ void NavEKF2_core::FuseMagnetometer()
...
+            const float correction_limit_ga = frontend->_mag_learn_limit*0.001*3;
```
rabad-matternet, line 397: What is our style on constants??  To someone who is not familiar with the code, 0.001 and 3 have no real meaning to me.  Should these be #defined, const, or constexpr instead of just some hardcoded number??

libraries/AP_NavEKF2/AP_NavEKF2_core.cpp
```diff
@@ -1440,9 +1440,11 @@ void NavEKF2_core::MagOffsetConstrain(void)
 {
     if (frontend->_mag_learn_limit > 0) {
         float limit_ga = frontend->_mag_learn_limit * 0.001f;
+        // allow higher limit on Z due to more variability observed in vehicles
+        float limit_ga_z = limit_ga * 3;
```
rabad-matternet, line 1444: Should 3 be #defined, const, or constexpr??

libraries/AP_NavEKF2/AP_NavEKF2_MagFusion.cpp
```diff
@@ -390,6 +391,17 @@ void NavEKF2_core::FuseMagnetometer()
...
+            correction = constrain_float(correction,
+                                         -correction_limit_ga, correction_limit_ga);
+            stateStruct.body_magfield[2] += correction;
+            gcs().send_text(MAV_SEVERITY_INFO, "EKF2 IMU%u magZ correction %.0f",(unsigned)imu_index, correction*1000);
```
rabad-matternet, line 401: should 1000 be #defined, const, or constexpr

libraries/AP_NavEKF2/AP_NavEKF2_MagFusion.cpp
```diff
@@ -390,6 +391,17 @@ void NavEKF2_core::FuseMagnetometer()
...
+            MagPred[2] += correction;
```
rabad-matternet, line 402: Same comment as above (line 396)

libraries/AP_NavEKF2/AP_NavEKF2_MagFusion.cpp
```diff
@@ -390,6 +391,17 @@ void NavEKF2_core::FuseMagnetometer()
         MagPred[1] = DCM[1][0]*magN + DCM[1][1]*magE  + DCM[1][2]*magD + magYbias;
         MagPred[2] = DCM[2][0]*magN + DCM[2][1]*magE  + DCM[2][2]*magD + magZbias;
 
+        if (pending_z_mag_correction && frontend->_mag_learn_limit > 0) {
+            pending_z_mag_correction = false;
+            float correction = magDataDelayed.mag[2] - MagPred[2];
+            const float correction_limit_ga = frontend->_mag_learn_limit*0.001*3;
+            correction = constrain_float(correction,
+                                         -correction_limit_ga, correction_limit_ga);
+            stateStruct.body_magfield[2] += correction;
```
rabad-matternet, line 400: Same comment above (line 396)

adriennet823 (approved):
Approving, though curious to know if @rabad-matternet 's questions will be addressed in this PR. I do like the idea of cleaning up some of the hard-coded values and creating some constexprs to improve readability and decrease possibility of errors

zalice (approved):
Echoing @rabad-matternet 's comments, at least for the magic numbers if the vector index refactor would be cumbersome.

Aksellence (approved):
Looks good to me. I would echo @rabad-matternet . Seems like a good point, it would definitely make the code more readable if we add constants, specially the unit change and the vector values. Other than that looks good and can't wait to see the test results. Thanks!

libraries/AP_NavEKF2/AP_NavEKF2_MagFusion.cpp
tridge: they are X, Y and Z axis, yes, used throughout the code base

libraries/AP_NavEKF2/AP_NavEKF2_MagFusion.cpp
tridge: 0.001 is for Gauss to milliGauss, 3 is the higher limit we are applying

rabad-matternet: Thank you for the explanation.

But the issue is readability.  Magic numbers should have some kind of "constant" rather than just hardcoded.

tridge: no, sorry. The code style we use in matternet must match that used in upstream ardupilot.
see my comment on GNC channel

libraries/AP_NavEKF2/AP_NavEKF2_core.cpp
tridge: see my GNC channel comment

libraries/AP_NavEKF2/AP_NavEKF2_MagFusion.cpp
tridge: see my GNC comment

tridge:
see my comments on code style on GNC channel

Aksellence (approved):

zalice:
Changes merged in https://github.com/matternet/ardupilot/pull/134

PR #122, opened 2020-07-28
DataFlash: use LOG_FILE_DSMROT even in log persistance window

tridge:
if we re-arm within 60s persist window then start new log if
LOG_FILE_DSMROT=1


zalice (approved):

adriennet823 (approved):

Aksellence (approved):
Looks good. Thanks!

Aksellence (approved):

PR #124, opened 2020-08-27
Send NAMED_VALUE_INT GPSPRIMARY at 1Hz

tridge:
This sends the mavlink msg NAMED_VALUE_INT with name "GPSPRIMARY" at 1Hz to indicate which GPS is being used
Values are:
 - 0: first GPS is primary
 - 1: 2nd GPS is primary
 - 2: blended GPS is primary


adriennet823 (approved):

Aksellence (approved):
Looks good.

zalice:
@tridge if both GPS units fail, will the reported primary GPS be the last GPS that was primary before failure?

zalice (approved):
Passed ground test.

PR #125, opened 2020-08-27
Fix issue with RTK GPS accuracy causing EKF divergence

tridge:
This fixes an issue found by a user flying with u-blox F9P GPS in a later version of ArduPilot. The issue arises when the GPS reports an accuracy of 1cm for a long period of time with RTK lock. 
The EKF variances associated with position can grow small enough that the EKF becomes unstable and diverges, resulting in an EKF failure.
The fix is to apply the minimum accuracy on GPS measurements at the point the measurement is taken instead of at the point the data is used. This ensures that all code paths use the data with an appropriate accuracy.
While this issue is very rare, the consequences of the issue happening are severe, with loss of EKF positioning. The fix is very simple.
The original PR is here:
https://github.com/ArduPilot/ardupilot/pull/15109


adriennet823 (approved):
Awesome, thanks for including the link to the original PR to provide more insight.

Aksellence (approved):
Looks good

zalice (approved):

PR #126, opened 2020-09-02
AP_Mission: allow mission clear while disarmed with mission running

tridge:
allows clear in AUTO while landed


zalice (approved):

adriennet823:
This can be closed, found that there is no need for this. @bhiln can you confirm?

bhiln:
> This can be closed, found that there is no need for this. @bhiln can you confirm?

confirmed, action taken in MATE

PR #127, opened 2020-09-17
Fixed EKF handling of compass failure and battery change

tridge:
This fixes two issues:
 - when a compass failed (stopped providing samples) we did not switch to the backup compass. We got away with this as the EKF, once converged, is able to fly for quite a long time without a compass.
 - when we change batteries between flights without rebooting the magnetic field of the vehicle could change. We should reset the mag body state variances when we are on the ground to ensure we can rapidly the changed field

Testing has been done in SITL with a test that flies two missions back to back without rebooting. The sequence is:
 - takeoff with deliberate X mag error of 100mGauss, fly a mission and land
 - change X mag offsets in SITL to zero and takeoff again. The EKF should learn the new field rapidly
 - part way through the 2nd flight fail the primary compass. The backup compass should be used, and it's offsets learned rapidly

Test logs before and after the fixes are here:
http://uav.tridgell.net/EKF/mag_offset_learn_sitl/

Without the fix we see this:
image
the initial offset learning is fine, and quickly approaches the correct value of 100mGauss. The learning on the 2nd flight is slow, and when the compass fails it stops learning altogether
with the fixes we get:
image
this shows rapidly learning in both flights, and correct handling of the compass failure

### DIFF
--- a/Tools/scripts/runcoptertest.py
+++ b/Tools/scripts/runcoptertest.py
@@ -7,9 +7,9 @@ def wait_heartbeat(mav, timeout=10):
     '''wait for a heartbeat'''
     start_time = time.time()
     while time.time() < start_time+timeout:
-        if mav.recv_match(type='HEARTBEAT', blocking=True, timeout=0.5) is not None:
+        if mav.recv_match(type='HEARTBEAT', blocking=True, timeout=2) is not None:
             return
-    failure("Failed to get heartbeat")    
+    print("Failed to get heartbeat")    
 
 def wait_mode(mav, modes, timeout=10):
     '''wait for one of a set of flight modes'''
@@ -35,25 +35,40 @@ def wait_time(mav, simtime):
         if t2 - t1 > simtime:
             break
 
-cmd = '../../Tools/autotest/sim_vehicle.py -D -S10'
+cmd = '../Tools/autotest/sim_vehicle.py -D -S10'
 mavproxy = pexpect.spawn(cmd, logfile=sys.stdout, timeout=30)
-mavproxy.expect("Ready to FLY")
+time.sleep(3)
 
-mav = mavutil.mavlink_connection('127.0.0.1:14550')
+mav = mavutil.mavlink_connection('tcp:127.0.0.1:5762')
 
-wait_time(mav, 2)
+mavproxy.expect("using GPS")
+mavproxy.expect("using GPS")
 mavproxy.send('mode GUIDED\n')
-mavproxy.send('param set SIM_SPEEDUP 5\n')
 wait_mode(mav, ['GUIDED'])
-mavproxy.expect('is using GPS')
-mavproxy.expect('is using GPS')
+mavproxy.send('param set SIM_MAG_OFS_X 0\n')
 mavproxy.send('arm throttle\n')
-mavproxy.expect('ARMED')
-mavproxy.send('mode AUTO\n')
-wait_mode(mav, ['AUTO'])
-wait_time(mav, 15)
+mavproxy.expect("Arming motors")
 mavproxy.send('long MISSION_START\n')
-mavproxy.send('module load console\n')
-mavproxy.send('module load map\n')
+#mavproxy.expect("Mission: 6 Jump")
+#mavproxy.send('param set SIM_MAG1_FAIL 1\n')
+mavproxy.expect("Disarming motors")
+
+#mavproxy.send('param set SIM_MAG1_FAIL 1\n')
+#mavproxy.expect("switching to compass")
+#mavproxy.expect("switching to compass")
+
+mavproxy.send('param set SIM_MAG_OFS_X 100\n')
+mavproxy.send('mode GUIDED\n')
+wait_mode(mav, ['GUIDED'])
+mavproxy.send('arm throttle force\n')
+mavproxy.expect("Arming motors")
+mavproxy.send('long MISSION_START\n')
+mavproxy.expect("Reached command #5")
+mavproxy.send("param set SIM_MAG_FAIL_MSK 1\n")
+#mavproxy.send('module load console\n')
+#mavproxy.send('module load map\n')
+#mavproxy.send('map set showsimpos 1\n')
+#mavproxy.logfile = None
+#mavproxy.interact()
+mavproxy.expect("Disarming motors")
 mavproxy.logfile = None
-mavproxy.interact()
--- a/libraries/AP_Compass/AP_Compass_SITL.cpp
+++ b/libraries/AP_Compass/AP_Compass_SITL.cpp
@@ -22,7 +22,13 @@ AP_Compass_SITL::AP_Compass_SITL(Compass &compass):
             // save so the compass always comes up configured in SITL
             save_dev_id(_compass_instance[i]);
         }
-        
+
+        // we want to simulate a calibrated compass by default, so set
+        // scale to 1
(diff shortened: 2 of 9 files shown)


libraries/AP_NavEKF2/AP_NavEKF2_Control.cpp
```diff
@@ -120,7 +121,7 @@ void NavEKF2_core::setWindMagStateLearningMode()
             P[21][21] = bodyMagFieldVar.z;
         } else {
             // set the variances equal to the observation variances
-            for (uint8_t index=18; index<=21; index++) {
+            for (uint8_t index=16; index<=21; index++) {
```
adriennet823, line 124: where did the start index shift from 18->16 come from?

libraries/AP_Compass/AP_Compass_SITL.cpp
```diff
@@ -123,19 +134,19 @@ void AP_Compass_SITL::_timer()
...
     }
 
-    _accum_count++;
-    if (_accum_count == 10) {
-        for (uint8_t i=0; i<SITL_NUM_COMPASSES; i++) {
+    for (uint8_t i=0; i<SITL_NUM_COMPASSES; i++) {
+        if (_accum_count[i] == 10) {
             _mag_accum[i] /= 2;
+            _accum_count[i] = 5;
+            _has_sample = true;
```
adriennet823, line 148: _mag_accum[i] *= 0.5 for performance?
Also what is the significance of resetting each value to 5? Can this be changed to a descriptively-named variable? (trying to minimize introducing more magic numbers)

adriennet823 (approved):
A few comments about "magic numbers"
Noticed a few square matrices using magic numbers as well (e.g. 19x19...21x21, etc.). I'm not aware of the motivation for these specific memory allocation sizes so I'm trying to maximize code readability. If possible could any of these be defined?

AydenChubbic (approved):

libraries/AP_NavEKF2/AP_NavEKF2_Control.cpp
tridge: it is a bug fix. See the definition of struct state_elements

libraries/AP_Compass/AP_Compass_SITL.cpp
tridge: this is a simulation change. It is a way of preventing overflow of accmulation counters. Normally accum_count stays small, if we stop reading samples for a while we need to prevent overflow.

tridge:
> Noticed a few square matrices using magic numbers as well (e.g. 19x19...21x21, etc.). I'm not aware of the motivation for these specific memory allocation sizes so I'm trying to maximize code readability. If possible could any of these be defined?

no, we don't want to change this in matternet branch. I completely agree that these numbers make the code hard to read, but that is how Paul coded it, and it is how his Matlab code generator spits out the fusion code. 
I have started on some work in ArduPilot master to fix this:
https://github.com/ArduPilot/ardupilot/pull/15110
but it is a large and complex change. We really want the matternet code to track the upstream code as closely as possible or we do not benefit from the testing done by the ardupilot community.
In matternet patches I follow the code style used in upstream ArduPilot. So comments on code style are not really useful unless they are in matternet specific code. This PR and most other PRs are not matternet specific.
Once I get the EKF to use better indexing in ArduPilot master that will flow through to matternet as matternet (slowly!) moves to newer versions of ardupilot.
If you'd like to understand the indexing in the EKF then you must look carefully at the definition of the state_elements structure. It is not possible to understand patches to the EKF code without knowing that structure.

libraries/AP_NavEKF2/AP_NavEKF2_Measurements.cpp
```diff
@@ -180,6 +180,54 @@ void NavEKF2_core::writeOptFlowMeas(uint8_t &rawFlowQuality, Vector2f &rawFlowRa
...
+    for (uint8_t i=1; i<maxCount; i++) {
+        uint8_t tempIndex = magSelectIndex + i;
+        // loop back to the start index if we have exceeded the bounds
+        if (tempIndex >= maxCount) {
+            tempIndex -= maxCount;
+        }
+        // if the magnetometer is allowed to be used for yaw and has a different index, we start using it
+        if (compass->healthy(tempIndex) && compass->use_for_yaw(tempIndex) && tempIndex != magSelectIndex) {
+            magSelectIndex = tempIndex;
+            gcs().send_text(MAV_SEVERITY_INFO, "EKF2 IMU%u switching to compass %u",(unsigned)imu_index,magSelectIndex);
```
zalice: Can we increase the severity to ALERT? @AkshayMatternet

adriennet823:
> If you'd like to understand the indexing in the EKF then you must look carefully at the definition of the state_elements structure. It is not possible to understand patches to the EKF code without knowing that structure.

Thanks @tridge ! I looked through the `state_elements` structure and that helped a lot

libraries/AP_NavEKF2/AP_NavEKF2_Measurements.cpp
Aksellence: Done!

Aksellence (approved):

zalice (approved):

Tools/scripts/runcoptertest.py
```diff
@@ -35,25 +35,40 @@ def wait_time(mav, simtime):
...
 mavproxy = pexpect.spawn(cmd, logfile=sys.stdout, timeout=30)
-mavproxy.expect("Ready to FLY")
+time.sleep(3)
 
-mav = mavutil.mavlink_connection('127.0.0.1:14550')
+mav = mavutil.mavlink_connection('tcp:127.0.0.1:5762')
 
-wait_time(mav, 2)
+mavproxy.expect("using GPS")
+mavproxy.expect("using GPS")
```
bhiln, line 45: Do you need to perform this expect twice?

Tools/scripts/runcoptertest.py
```diff
@@ -35,25 +35,40 @@ def wait_time(mav, simtime):
...
 mavproxy.send('long MISSION_START\n')
-mavproxy.send('module load console\n')
-mavproxy.send('module load map\n')
+#mavproxy.expect("Mission: 6 Jump")
+#mavproxy.send('param set SIM_MAG1_FAIL 1\n')
+mavproxy.expect("Disarming motors")
+
+#mavproxy.send('param set SIM_MAG1_FAIL 1\n')
+#mavproxy.expect("switching to compass")
+#mavproxy.expect("switching to compass")
```
bhiln, line 58: If this commented code isn't used anymore, please remove

Tools/scripts/runcoptertest.py
```diff
@@ -35,25 +35,40 @@ def wait_time(mav, simtime):
...
+mavproxy.send('arm throttle force\n')
+mavproxy.expect("Arming motors")
+mavproxy.send('long MISSION_START\n')
+mavproxy.expect("Reached command #5")
+mavproxy.send("param set SIM_MAG_FAIL_MSK 1\n")
+#mavproxy.send('module load console\n')
+#mavproxy.send('module load map\n')
+#mavproxy.send('map set showsimpos 1\n')
+#mavproxy.logfile = None
+#mavproxy.interact()
```
bhiln, line 72: Same with this commented code

zalice:
Merged in https://github.com/matternet/ardupilot/pull/134

PR #128, opened 2020-09-25
GPS switching logic bugfix

Aksellence:
This build fixes the issue with the GPS switching logic, where the aircraft, when Armed, will not switch to the better GPS, if available. This missing case can cause trouble in rare edge cases where the currently chosen primary GPS tends to be showing a degraded performance, while the vehicle is in the air.


libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -789,7 +789,13 @@ void AP_GPS::update(void)
...
                     }
                     
-                    else if (status_i == status_primary && !hal.util->get_soft_armed() && state[i].num_sats >= state[primary_instance].num_sats+3) {
+                    // When vehicle is Disarmed, and current primary GPS has less than 3 satellites, switch to the other available GPS.
+                    else if (!hal.util->get_soft_armed() && status_i == status_primary && state[i].num_sats >= state[primary_instance].num_sats+3) {
+                        should_switch = true;                        
+                    }
+
+                    // When vehicle is Armed, and current primary GPS has less than 6 satellites, switch to the other available GPS.
+                    else if (!hal.util->get_soft_armed() && status_i == status_primary && state[i].num_sats >= state[primary_instance].num_sats+6) {
```
zalice: This checks if disarmed rather than armed

zalice:
Is this PR mixed with the mag Z offsets PR?
Can we merge the magZ PR since it passed flight test? That will also simplify this review.

libraries/AP_GPS/AP_GPS.cpp
Aksellence: Yes. Thanks @zalice! Resolved!

Aksellence:
> Is this PR mixed with the mag Z offsets PR?
> Can we merge the magZ PR since it passed the flight test? That will also simplify this review.

Yes it is.

AydenChubbic (approved):

Tools/scripts/runcoptertest.py
```diff
@@ -35,25 +35,40 @@ def wait_time(mav, simtime):
...
 mavproxy = pexpect.spawn(cmd, logfile=sys.stdout, timeout=30)
-mavproxy.expect("Ready to FLY")
+time.sleep(3)
 
-mav = mavutil.mavlink_connection('127.0.0.1:14550')
+mav = mavutil.mavlink_connection('tcp:127.0.0.1:5762')
 
-wait_time(mav, 2)
+mavproxy.expect("using GPS")
+mavproxy.expect("using GPS")
```
bhiln, line 45: Looks like L44 and L45 are duplicates, is this intentional?

Tools/scripts/runcoptertest.py
```diff
@@ -35,25 +35,40 @@ def wait_time(mav, simtime):
...
+mavproxy.send('mode GUIDED\n')
+wait_mode(mav, ['GUIDED'])
+mavproxy.send('arm throttle force\n')
+mavproxy.expect("Arming motors")
+mavproxy.send('long MISSION_START\n')
+mavproxy.expect("Reached command #5")
+mavproxy.send("param set SIM_MAG_FAIL_MSK 1\n")
+#mavproxy.send('module load console\n')
+#mavproxy.send('module load map\n')
+#mavproxy.send('map set showsimpos 1\n')
```
bhiln, line 70: please remove dead code

libraries/AP_NavEKF2/AP_NavEKF2_Measurements.cpp
```diff
@@ -180,6 +180,54 @@ void NavEKF2_core::writeOptFlowMeas(uint8_t &rawFlowQuality, Vector2f &rawFlowRa
...
+    P[19][19] = sq(frontend->_magNoise);
+    P[20][20] = P[19][19];
+    P[21][21] = P[19][19];
+}
+
+// try changing to another compass
+void NavEKF2_core::tryChangeCompass(void)
+{
+    const auto *compass = _ahrs->get_compass();
+    const uint8_t maxCount = compass->get_count();
```
bhiln, line 197: Consider changing maxCount name to reflect it's count of magnetometers: `macCountMagnetomers`

libraries/SITL/SITL.cpp
```diff
@@ -159,6 +159,7 @@ const AP_Param::GroupInfo SITL::var_info2[] = {
     AP_GROUPINFO("GPS_DRIFT_S", 35, SITL,  gps_drift_spd, 0),
 
     AP_GROUPINFO("MAG_SCALING",    60, SITL,  mag_scaling, 1),
+    AP_GROUPINFO("MAG_FAIL_MSK",  61, SITL,  mag_fail_mask, 0),
```
bhiln, line 162: nitpick: do yo mind aligning the arguments with L161? thanks!

bhiln (approved):

libraries/AP_Compass/AP_Compass_SITL.cpp
```diff
@@ -123,19 +134,19 @@ void AP_Compass_SITL::_timer()
         correct_field(f, _compass_instance[i]);
 
         _mag_accum[i] += f;
+        _accum_count[i]++;
     }
 
     if (!_sem->take(HAL_SEMAPHORE_BLOCK_FOREVER)) {
         return;
     }
```
rabad-matternet, line 142: I know this isn't part of your changes but just something that caught my eye...

If this call to take the semaphore will wait forever, what would make take() return false??

libraries/AP_NavEKF2/AP_NavEKF2_Control.cpp
```diff
@@ -120,7 +121,7 @@ void NavEKF2_core::setWindMagStateLearningMode()
             P[21][21] = bodyMagFieldVar.z;
         } else {
             // set the variances equal to the observation variances
-            for (uint8_t index=18; index<=21; index++) {
+            for (uint8_t index=16; index<=21; index++) {
                 P[index][index] = sq(frontend->_magNoise);
```
rabad-matternet, line 125: What are the significances of indices 16 and 21?  Can these magic numbers be #define, const, or const expr??

libraries/AP_NavEKF2/AP_NavEKF2_Measurements.cpp
```diff
@@ -180,6 +180,54 @@ void NavEKF2_core::writeOptFlowMeas(uint8_t &rawFlowQuality, Vector2f &rawFlowRa
...
+// reset variances for mag body state
+void NavEKF2_core::resetMagBodyVariances(void)
+{
+    zeroCols(P,19,21);
+    zeroRows(P,19,21);
+    P[19][19] = sq(frontend->_magNoise);
+    P[20][20] = P[19][19];
+    P[21][21] = P[19][19];
+}
+
```
rabad-matternet, line 192: What is significant about 19, 20, and 21?  Can these magic numbers be #define, const, or constexpr?

libraries/AP_NavEKF2/AP_NavEKF2_Control.cpp
rabad-matternet: Or at least put a comment explaining why we're going from 16 to 21?

libraries/AP_NavEKF2/AP_NavEKF2_Measurements.cpp
rabad-matternet: Or at least put a comment explaining 19 - 21?

libraries/AP_NavEKF2/AP_NavEKF2_Measurements.cpp
```diff
@@ -180,6 +180,54 @@ void NavEKF2_core::writeOptFlowMeas(uint8_t &rawFlowQuality, Vector2f &rawFlowRa
...
+// try changing to another compass
+void NavEKF2_core::tryChangeCompass(void)
```
adriennet823, line 194: Just a nitpick, but to keep wording uniform throughout the code and to avoid confusion, I'd recommend renaming this function to `tryCompassSwitch()` or `trySwitchCompass()` since everywhere else we see flags like should_switch or the text below says "switching to compass %u" and the PR itself uses "switch" rather than "change." I think it's also more descriptive of the behavior since change can sometimes mean a replacement whereas switch usually implies that we are moving back and forth between existing hardware

libraries/AP_NavEKF2/AP_NavEKF2_Measurements.cpp
```diff
@@ -180,6 +180,54 @@ void NavEKF2_core::writeOptFlowMeas(uint8_t &rawFlowQuality, Vector2f &rawFlowRa
...
+            // zero the learned magnetometer bias states
+            stateStruct.body_magfield.zero();
+            // clear the measurement buffer
+            storedMag.reset();
+            // clear the data waiting flag so that we do not use any data pending from the previous sensor
+            magDataToFuse = false;
+            // request a reset of the magnetic field states
+            magStateResetRequest = true;
+            // declare the field unlearned so that the reset request will be obeyed
+            magFieldLearned = false;
```
adriennet823, line 222: similar to the `resetMagBodyVariances()` call below, can all of this be moved to a function like `resetMagFlags()` or `resetMagStates()`?

adriennet823 (approved):
Some minor requests, non-blocking

PR #129, opened 2020-10-06
AP_Arming: Fixed prearm alert message

Aksellence:
Empty pre arm alert message being generated by the Ardupilot. It was fixed by changing the pre-arm message such that it is more descriptive. More information can be found here:
https://matternet.atlassian.net/browse/NCT-1025
https://matternet.atlassian.net/browse/APM-62

### DIFF
--- a/libraries/AP_Arming/AP_Arming.cpp
+++ b/libraries/AP_Arming/AP_Arming.cpp
@@ -145,7 +145,7 @@ void AP_Arming::check_failed(const enum AP_Arming::ArmingChecks check, bool repo
         return;
     }
     char taggedfmt[MAVLINK_MSG_STATUSTEXT_FIELD_TEXT_LEN+1];
-    hal.util->snprintf((char*)taggedfmt, sizeof(taggedfmt)-1, "PreArm: %s", fmt);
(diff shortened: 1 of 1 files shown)


libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -145,7 +145,7 @@ void AP_Arming::check_failed(const enum AP_Arming::ArmingChecks check, bool repo
         return;
     }
     char taggedfmt[MAVLINK_MSG_STATUSTEXT_FIELD_TEXT_LEN+1];
-    hal.util->snprintf((char*)taggedfmt, sizeof(taggedfmt)-1, "PreArm: %s", fmt);
+    hal.util->snprintf((char*)taggedfmt, sizeof(taggedfmt)-1, "PreArm check(s) failed");
```
zalice, line 148: @AkshayMatternet which failing checks do not populate the alert content? Are there checks that populated the fmt? This change appears to hardcode the alert message rather than make it dynamically populate with the failing prearm check

adriennet823: How can we see which prearm checks actually failed?

zalice:
Merged in another PR

PR #131, opened 2020-10-16
AC_AttitudeControl: added WPNAV_XTRACK_MAX parameter

tridge:
this allows the leash lateral error (the crosstrack) to be limited, which can help constrain
the path close to the desired waypoints in high wind

here is a SITL test with no limit:
image

and here is the same wind simulation with WPNAV_XTRACK_MAX=30
image

### DIFF
--- a/libraries/AC_WPNav/AC_WPNav.cpp
+++ b/libraries/AC_WPNav/AC_WPNav.cpp
@@ -67,6 +67,14 @@ const AP_Param::GroupInfo AC_WPNav::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("RFND_USE",   10, AC_WPNav, _rangefinder_use, 1),
 
+    // @Param: XTRACK_MAX
+    // @DisplayName: Waypoint crosstrack maximum
+    // @Description: This controls the maximum crosstrack distance in waypoint missions for fast waypoints. A value of zero means no limit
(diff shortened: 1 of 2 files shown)


libraries/AC_WPNav/AC_WPNav.cpp
```diff
@@ -379,6 +387,16 @@ bool AC_WPNav::advance_wp_target_along_track(float dt)
         reached_leash_limit = true;
     }
 
+    /*
+      also check for crosstrack limit, allowing us to constrain leash
+      point to within a given distance of the track. This can be used
+      to prevent high wind from causing a blowout of the distance from
+      the desired track
+     */
+    if (_crosstrack_max > 0 && _track_error_xy*0.01 > _crosstrack_max) {
```
zalice, line 396: @tridge does the `_crosstrack_max > 0` require the use of machine epsilon in the comparison? I see that it's defined as `AP_Float` -- how is that different from vanilla `float`?

tridge: using > 0 here just means it has been set at all. We could use is_positive(), but not really needed
AP_Float is a float that is a parameter in AP_Param

adriennet823 (approved):

Aksellence (approved):

zalice (approved):

PR #135, opened 2020-11-03
AP_GPS: Added GPS switching logic for the (missing) case when vehicle is armed

Aksellence:
This build fixes the issue with the GPS switching logic, where the aircraft, when Armed, will not switch to the better GPS, if available. This missing case can cause trouble in rare edge cases where the currently chosen primary GPS tends to be showing a degraded performance, while the vehicle is in the air.
This PR is already reviewed before but made a new one, as the old PR had some other changes which are already merged in the master.  This PR is on top of the latest Ardupilot release (mttr-1.0.14).

### DIFF
--- a/libraries/AP_GPS/AP_GPS.cpp
+++ b/libraries/AP_GPS/AP_GPS.cpp
@@ -789,7 +789,13 @@ void AP_GPS::update(void)
                         should_switch = true;
                     }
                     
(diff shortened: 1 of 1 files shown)


adriennet823 (approved):

AydenChubbic (approved):

bhiln (approved):

zalice:
Switching logic will be refactored

PR #140, opened 2021-01-04
support USER_1 mission item as failsafe jump

tridge:
This does a conditional jump to a specified waypoint based on the time since the last GCS heartbeat. It is used to implement safe zones in the mission where landing sequences are conditionally started if the companion computer becomes unavailable
Here is an example mission:

image

in this mission WP6 is the custom USER_1 mission item. That mission item takes two parameters:
 - param1 is the target of the conditional jump. It must be a valid WP number. In this case it points at WP8, which is a NAV_LOITER_TURNS waypoint
 -  param2 is the number of seconds that we got the last heartbeat from the GCS for us to not do the jump. The minimum value is 2 (a value below 2 will use 2). In this case we're using a value of 3.0, which means that if we pass this USER_1 waypoint and we haven't had a heartbeat from the GCS in the last 3 seconds then we will do the jump to WP 8.

Unlike a DO_JUMP there is no counter associated with a USER_1 conditional jump.

Some possible changes:
 - we could use an otherwise unused field of DO_JUMP instead if you prefer to give the timeout, with zero giving normal DO_JUMP behaviour

I've discussed with the ArduPilot multicopter lead @rmackay9 about including this feature in standard ArduPilot, and he would prefer that a lua script be used for the 4.0 or 4.1 version.

### DIFF
--- a/ArduCopter/mode.h
+++ b/ArduCopter/mode.h
@@ -374,6 +374,7 @@ private:
 #endif
     void do_payload_place(const AP_Mission::Mission_Command& cmd);
     void do_RTL(void);
+    bool do_failsafe_jump(const AP_Mission::Mission_Command& cmd);
 
     bool verify_takeoff();
     bool verify_land();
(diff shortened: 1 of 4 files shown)


zalice (approved):

Aksellence (commented):
Looks good

PR #141, opened 2021-01-11
AP_Rangefinder: change lightware lost signal timer from 1 to 20

zalice:
In response to: https://github.com/ArduPilot/ardupilot/issues/15615

### DIFF
--- a/libraries/AP_RangeFinder/AP_RangeFinder_LightWareI2C.cpp
+++ b/libraries/AP_RangeFinder/AP_RangeFinder_LightWareI2C.cpp
@@ -274,7 +274,7 @@ bool AP_RangeFinder_LightWareI2C::sf20_init()
     // When it is supported the expected response would be "e:1".
 
     // Changes the number of lost signal confirmations: 1 [1..250].
-    if(!sf20_send_and_expect("#LC,1\r\n", "lc:1")) {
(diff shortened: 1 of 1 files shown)


adriennet823 (commented):
The linked issue also references this PR as part of the fix: https://github.com/ArduPilot/ardupilot/pull/15846 should this be added as well or would that be a separate PR?

zalice:
> The linked issue also references this PR as part of the fix: [ArduPilot/ardupilot#15846](https://github.com/ArduPilot/ardupilot/pull/15846) should this be added as well or would that be a separate PR?

I haven't dug enough through the 3.6.7 and 4.0 delta to understand whether we can use glitch protected alt in our 3.6.7 like this PR https://github.com/ArduPilot/ardupilot/pull/15846/files
@tridge is our 3.6.7 up to date on glitch protection? or is this something available in 4.0+?

Aksellence:
> The linked issue also references this PR as part of the fix: [ArduPilot/ardupilot#15846](https://github.com/ArduPilot/ardupilot/pull/15846) should this be added as well or would that be a separate PR?

@adriennet823 I looked into the repository and I can't find glitch-protected alt in our 3.6.7 branch but was surely able to find it in 4.0 branch. Also, I think if we are thinking of applying that change, can we apply that change in a separate PR, as this anyway seems to be beyond the scope of this PR?

Aksellence (approved):

PR #162, opened 2021-03-23
Feature/MATE-1579 Play tune upon mission load

adriennet823:
Removes play tune send message from MATE, which would intermittently drop. Adds new tune to `_tones` and plays tune when last mission item from GCS is received. Barrier in `play_tone` prevents tune from playing while armed.

### DIFF
--- a/libraries/AP_Notify/ToneAlarm.cpp
+++ b/libraries/AP_Notify/ToneAlarm.cpp
@@ -95,7 +95,6 @@ const AP_ToneAlarm::Tone AP_ToneAlarm::_tones[] {
     { "MNBGG", false },
 #define AP_NOTIFY_TONE_FLIGHT_PLAN_LOAD 31
     { "MFT220 ML O3ef O4c", true },
-"}
 };
 
 bool AP_ToneAlarm::init()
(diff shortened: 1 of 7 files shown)


AydenChubbic (approved):

libraries/AP_Notify/ToneAlarm.cpp
```diff
@@ -93,6 +93,8 @@ const AP_ToneAlarm::Tone AP_ToneAlarm::_tones[] {
     { "MFMST200L64O4ceceP32ceceP8df#df#P32df#df#P8L16gf#g>c", false },
 #define AP_NOTIFY_TONE_NO_SDCARD 30
     { "MNBGG", false },
+#define AP_NOTIFY_TONE_FLIGHT_PLAN_LOAD 31
+    { "MFT220 ML O3ef O4c", true },
```
tridge, line 97: the 'true' means continuous - is that really wanted? I would have thought a one-off play of the tune would be what you would want

tridge (changes requested):
This really should be squashed down to 2 commits, one "AP_Notify" and one "GCS_MAVLink" commit.

libraries/AP_Notify/ToneAlarm.cpp
adriennet823: I misinterpreted the meaning of "continuous" -- I will change this to false. One-off is what I want

adriennet823:
> This really should be squashed down to 2 commits, one "AP_Notify" and one "GCS_MAVLink" commit.

Good point, I was trying different solutions hence the commit history. I'll re-issue the PR with the 2 commits

PR #168, opened 2021-04-01
AP_GPS: fixed nullptr deref for dual-GPS (3.6 version)

tridge:
the patch for the special case for F9 handling could dereference
nullptr, possibly resulting in a watchdog
while we have not seen the issue with 3.6.x, we should have the protection in


zalice:
Since we proved that the nullptr deref is not possible in this version of 3.6, I will close this PR

PR #171, opened 2021-04-16
Added compass primary defaults (mttr-4.0)

Aksellence:
Added compass primary default for setting the CAN compass as primary and I2C compass as a secondary fallback heading reference sensor.


libraries/AP_HAL_ChibiOS/hwdef/CubeOrange/defaults.parm
```diff
@@ -13,3 +13,7 @@ SERIAL5_PROTOCOL 1
 # lower heater gains
 BRD_IMUHEAT_I 0.07
 BRD_IMUHEAT_P 50
+
+# compass primary defaults
+COMPASS_PRIO1_ID 73475
```
zalice, line 18: This should be
```
COMPASS_PRIO1_ID COMPASS_DEV_ID2
COMPASS_PRIO2_ID COMPASS_DEV_ID
```

zalice (approved):

zalice (changes requested):
Please see requested param values

libraries/AP_HAL_ChibiOS/hwdef/CubeOrange/defaults.parm
Aksellence: I don't think so. As per me, it should be the value associated with the particular compass_dev_id that we want at a certain priority and not the string mentioning the 'compass_dev_id<ID number>'

tridge: using the string won't work, as the parser doesn't know how to decode parameter names in the right hand column

libraries/AP_HAL_ChibiOS/hwdef/CubeOrange/defaults.parm
```diff
@@ -13,3 +13,7 @@ SERIAL5_PROTOCOL 1
...
+COMPASS_PRIO2_ID 592905
```
tridge: I'd also suggest setting COMPASS_PRIO3_ID to 0, to ensure it doesn't have a conflict

tridge (changes requested):

libraries/AP_HAL_ChibiOS/hwdef/CubeOrange/defaults.parm
zalice: Thank you for the clarification, @tridge !

zalice (approved):

libraries/AP_HAL_ChibiOS/hwdef/CubeOrange/defaults.parm
Aksellence: Done. Thanks!

zalice (approved):

PR #173, opened 2021-04-22
Allow Arming with RC failsafe when in GUIDED (mttr-rebase-4.0)

Aksellence:
Found an issue where the aircraft would not arm with RC radio Off, with the current FS configuration. The aircraft would need double arming, in order to arm. This PR comprises the Fix for both issues.
The SITL test results can be found [here](https://matternet.atlassian.net/wiki/spaces/APM/pages/1518207330/RC+Failsafe+SITL+testing+mttr-rebase-4.0+Cube+Orange
)


ArduCopter/events.cpp
```diff
@@ -34,7 +34,7 @@ void Copter::failsafe_radio_on_event()
             desired_action = Failsafe_Action_SmartRTL_Land;
             break;
         case FS_THR_ENABLED_ALWAYS_LAND:
-            desired_action = Failsafe_Action_Land;
+            desired_action = Failsafe_Action_None;
```
zalice: @AkshayMatternet what behavior issue does this change address? It's not note in the PR description

Aksellence: This stops the aircraft from triggering the failsafe action, which is, the aircraft not arming and switching to LAND mode every time it gets an Arm command while the aircraft is on the current FS configuration and the RC is turned Off.

ArduCopter/events.cpp
```diff
@@ -34,7 +34,12 @@ void Copter::failsafe_radio_on_event()
             desired_action = Failsafe_Action_SmartRTL_Land;
             break;
         case FS_THR_ENABLED_ALWAYS_LAND:
-            desired_action = Failsafe_Action_Land;
+            //    Do not switch to land in Auto mode, Switch to LAND in Manual mode
+            if (flightmode->in_guided_mode() || control_mode == Mode::Number::AUTO) {
```
tridge: this needs to also check the fs_options bit, so needs to add:
 && failsafe_option(FailsafeOption::RC_CONTINUE_IF_AUTO)

ArduCopter/events.cpp
```diff
@@ -355,7 +360,8 @@ bool Copter::should_disarm_on_failsafe() {
             return !ap.auto_armed && ap.land_complete;
         case Mode::Number::GUIDED:
             if (g.failsafe_throttle == FS_THR_ENABLED_CONTINUE_MISSION ||
-                g.failsafe_throttle == FS_THR_ENABLED_CONTINUE_MISSION_ALWAYS_LAND) {
+                g.failsafe_throttle == FS_THR_ENABLED_CONTINUE_MISSION_ALWAYS_LAND ||
+                g.failsafe_throttle == FS_THR_ENABLED_ALWAYS_LAND) {
```
tridge, line 367: this should check failsafe_option(FailsafeOption::RC_CONTINUE_IF_AUTO) as well

tridge (changes requested):

ArduCopter/events.cpp
Aksellence: Okay added, also added 'RC_CONTINUE_IF_GUIDED', as we would need the same behavior for Guided as well.

ArduCopter/events.cpp
Aksellence: Added, also added this for Guided, as we need the same behavior for Guided. Thanks!

ArduCopter/events.cpp
zalice: Why should this check `RC_CONTINUE_IF_AUTO` if the case entry requires that the aircraft is in GUIDED? Should this bit check be in the `AUTO` case?

Aksellence: It should be for both the modes, AUTO and GUIDED. So added it for AUTO as well as GUIDED

zalice (approved):

PR #174, opened 2021-04-22
AP_IRLock: added logging of raw IRlock I2C data

tridge:
implements feature for APM-66

Note that this logs every frame we get. We should look at whether the logging bandwidth is worth keeping this permanently or if we want to enable it with a config option.


zalice (approved):

zalice:
@tridge is there a case in which the logging of the targets could overwhelm the controller? e.g. should we limit the logging to only the first 10 targets in the list?

PR #177, opened 2021-06-07
Copter events.cpp: Remove FS_GCS_DISABLED check in RC failsafe

zalice:
Currently when the aircraft is armed in GUIDED mode without a RC powered on, the aircraft automatically disarms with the message "Radio Failsafe" due to the check for not FS_GCS_DISABLED failing since our configuration has GCS FS disabled.

This PR removes the FS_GCS_DISABLED requirement for RC failsafe disarm when the aircraft is armed in GUIDED mode.


tridge (approved):
looks good!

AydenChubbic (approved):

Aksellence (approved):
Looks good.

PR #179, opened 2021-06-11
AVX-283 Revert RNGFND1_MIN_CM in radar CAN TXR

rabad-matternet:
- reverted RNGFND1_MIN_CM to 5


adriennet823 (approved):

AydenChubbic (approved):

zalice (approved):
Can we do a version bump with this PR? So that we'll be able to tell which aircraft have the change

PR #180, opened 2021-06-11
AVX-284 Output SNR from radar

rabad-matternet:
- parsed radar data for SNR
- output SNR via overloaded "field_of_view" in uavcan.equipment.range_sensor.Measurement (id: 1050)
- limited SNR output to US-D1 radar

### DIFF
--- a/libraries/AP_Logger/LogStructure.h
+++ b/libraries/AP_Logger/LogStructure.h
@@ -1938,14 +1938,14 @@ struct PACKED log_PSC {
       "BAT", "QBfffffcf", "TimeUS,Instance,Volt,VoltR,Curr,CurrTot,EnrgTot,Temp,Res", "s#vvAiJOw", "F-000!/?0" },  \
     { LOG_CURRENT_CELLS_MSG, sizeof(log_Current_Cells), \
       "BCL", "QBfHHHHHHHHHHHH", "TimeUS,Instance,Volt,V1,V2,V3,V4,V5,V6,V7,V8,V9,V10,V11,V12", "s#vvvvvvvvvvvvv", "F-0CCCCCCCCCCCC" }, \
-	{ LOG_ATTITUDE_MSG, sizeof(log_Attitude),\
+	  { LOG_ATTITUDE_MSG, sizeof(log_Attitude),\
       "ATT", "QccccCCCCB", "TimeUS,DesRoll,Roll,DesPitch,Pitch,DesYaw,Yaw,ErrRP,ErrYaw,AEKF", "sddddhhdh-", "FBBBBBBBB-" }, \
     { LOG_MAG_MSG, sizeof(log_MAG), \
       "MAG", "QBhhhhhhhhhBI",    "TimeUS,I,MagX,MagY,MagZ,OfsX,OfsY,OfsZ,MOX,MOY,MOZ,Health,S", "s#GGGGGGGGG-s", "F-CCCCCCCCC-F" }, \
     { LOG_MODE_MSG, sizeof(log_Mode), \
       "MODE", "QMBB",         "TimeUS,Mode,ModeNum,Rsn", "s---", "F---" }, \
     { LOG_RFND_MSG, sizeof(log_RFND), \
-      "RFND", "QBCBB", "TimeUS,Instance,Dist,Snr,Stat,Orient", "s#m--", "F-B--" }, \
+      "RFND", "QBCCBB", "TimeUS,Instance,Dist,Snr,Stat,Orient", "s#m---", "F-B---" }, \
     { LOG_MAV_STATS, sizeof(log_MAV_Stats), \
       "DMS", "QIIIIBBBBBBBBB",         "TimeUS,N,Dp,RT,RS,Fa,Fmn,Fmx,Pa,Pmn,Pmx,Sa,Smn,Smx", "s-------------", "F-------------" }, \
     { LOG_BEACON_MSG, sizeof(log_Beacon), \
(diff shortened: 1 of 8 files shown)


libraries/AP_RangeFinder/AP_RangeFinder.cpp
```diff
@@ -658,6 +658,15 @@ AP_RangeFinder_Backend *RangeFinder::find_instance(enum Rotation orientation) co
     return nullptr;
 }
 
+uint16_t RangeFinder::snr(enum Rotation orientation) const
+{
+    AP_RangeFinder_Backend *backend = find_instance(orientation);
```
zalice, line 663: Should there be a check for rangefinder type here and return 0 if not ulanding / radar? Or am I missing where RangeFinder_State snr is initialized to 0?

zalice: Ah, I see it now. It's a pointer t o AP_RangeFinder_Backend which is an instance of uLanding. Nvm :)

zalice (approved):

libraries/AP_Logger/LogStructure.h
```diff
@@ -618,6 +618,7 @@ struct PACKED log_RFND {
     uint64_t time_us;
     uint8_t instance;
     uint16_t dist;
+    uint16_t snr;
```
zalice, line 621: Could you also please update
```
    { LOG_RFND_MSG, sizeof(log_RFND), \      "RFND", "QBCBB", "TimeUS,Instance,Dist,Stat,Orient", "s#m--", "F-B--" }, \
```
to include snr? L1408 of this file

rabad-matternet: Done. (It was line ~1946 in this branch)

zalice: Thank you! I was looking to apply this commit to our cube firmware and must have been looking at the line numbers in there 🙈

libraries/AP_Logger/LogStructure.h
```diff
@@ -1943,7 +1945,7 @@ struct PACKED log_PSC {
     { LOG_MODE_MSG, sizeof(log_Mode), \
       "MODE", "QMBB",         "TimeUS,Mode,ModeNum,Rsn", "s---", "F---" }, \
     { LOG_RFND_MSG, sizeof(log_RFND), \
-      "RFND", "QBCBB", "TimeUS,Instance,Dist,Stat,Orient", "s#m--", "F-B--" }, \
+      "RFND", "QBCBB", "TimeUS,Instance,Dist,Snr,Stat,Orient", "s#m--", "F-B--" }, \
```
zalice: We need to also add to "QBCBB" and the format strings. Since it's the same type as dist, can you use that as reference? Meaning find the index of Dist in each of those and add another of that character to the string.
For QBCBB --> QBCCBB

zalice: s#m-- to s#mm--

zalice (changes requested):
Could you please update the logging format strings?

libraries/AP_Logger/LogStructure.h
rabad-matternet: Done.  See next commit

rabad-matternet: > s#m-- to s#mm--

this becomes s#m---

'm' means meters.  SNR is in decibels (usually)

libraries/AP_Logger/LogStructure.h
```diff
@@ -1936,14 +1938,14 @@ struct PACKED log_PSC {
...
-	{ LOG_ATTITUDE_MSG, sizeof(log_Attitude),\
+	  { LOG_ATTITUDE_MSG, sizeof(log_Attitude),\
       "ATT", "QccccCCCCB", "TimeUS,DesRoll,Roll,DesPitch,Pitch,DesYaw,Yaw,ErrRP,ErrYaw,AEKF", "sddddhhdh-", "FBBBBBBBB-" }, \
     { LOG_MAG_MSG, sizeof(log_MAG), \
       "MAG", "QBhhhhhhhhhBI",    "TimeUS,I,MagX,MagY,MagZ,OfsX,OfsY,OfsZ,MOX,MOY,MOZ,Health,S", "s#GGGGGGGGG-s", "F-CCCCCCCCC-F" }, \
     { LOG_MODE_MSG, sizeof(log_Mode), \
       "MODE", "QMBB",         "TimeUS,Mode,ModeNum,Rsn", "s---", "F---" }, \
     { LOG_RFND_MSG, sizeof(log_RFND), \
-      "RFND", "QBCBB", "TimeUS,Instance,Dist,Stat,Orient", "s#m--", "F-B--" }, \
+      "RFND", "QBCCBB", "TimeUS,Instance,Dist,Snr,Stat,Orient", "s#m---", "F-B---" }, \
```
zalice, line 1948: Should SNR be m and B for the two format messages? Rather than - ?

zalice (commented):
Left a question about the log format

libraries/AP_Logger/LogStructure.h
rabad-matternet: > Should SNR be m and B for the two format messages? Rather than - ?

`QBCCBB` specifies the type. C is a uint16_t.
`s#m---` specifies units. '-' means no units.  'm' means meters (or metres, for those crazy commonwealth peeps lol).  Meters is not applicable to SNR.  There is no representation for dB.
`F-B---` specifies multiplier.  '-' means no mulitplier.

https://github.com/matternet/ardupilot/tree/master_can_txr/libraries/AP_Logger#logger-notes

zalice: Thanks for the in depth explanation and link 🙏

zalice (approved):

PR #183, opened 2021-07-09
AP_GPS Add configurable GPS difference and min sats for arming

zalice:
Added configuration options:
- POS_DIFF_MAX
- NUM_SATS_MIN

Also require enhanced GPS fix for arming.

### DIFF
--- a/libraries/AP_GPS/AP_GPS.cpp
+++ b/libraries/AP_GPS/AP_GPS.cpp
@@ -298,6 +298,24 @@ const AP_Param::GroupInfo AP_GPS::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("DRV_OPTIONS", 22, AP_GPS, _driver_options, 0),
 
+    // @Param: MAX_POS_DIFF
+    // @DisplayName: Maximum Position Difference Between GPSs Accepted
+    // @Description: Sets the maximum allowed difference in meters allowed in GPS positions before takeoff.
+    // @Units: m
+    // @Range: 1.0 50.0
+    // @User: Advanced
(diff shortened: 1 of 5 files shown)


AydenChubbic (approved):

Aksellence (approved):
Looks good. Thanks for adding this @zalice

libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -298,6 +298,24 @@ const AP_Param::GroupInfo AP_GPS::var_info[] = {
...
     AP_GROUPINFO("DRV_OPTIONS", 22, AP_GPS, _driver_options, 0),
 
+    // @Param: MAX_POS_DIFF
+    // @DisplayName: Maximum Position Difference Between GPSs Accepted
+    // @Description: Sets the maximum allowed difference in meters allowed in GPS positions before takeoff.
+    // @Units: m
+    // @Range: 1.0 50.0
+    // @User: Advanced
+    // @RebootRequired: True
+    AP_GROUPINFO("MAX_POS_DIFF", 23, AP_GPS, _max_pos_diff, 5.0f),
```
tridge: I suggest using indexes 62 and 63 so there is a big gap from the standard indexes. This will save you pain when this is rebased on 4.1 in the future as the IDs won't conflict

tridge (changes requested):

libraries/AP_GPS/AP_GPS.cpp
zalice: Thank you for noting this @tridge. I pushed this change

PR #186, opened 2021-07-14
AVX-323 Enable hardware flow control on Serial 1 (cube orange)

rabad-matternet:
- changed flow control param from auto to enable


bhiln (approved):

libraries/AP_HAL_ChibiOS/hwdef/MttrCubeOrange/defaults.parm
```diff
@@ -34,7 +34,7 @@ BATT_MONITOR	20
 BATT_VOLT_MULT	22.7
 BRD_SAFETYENABLE	0
 BRD_SAFETY_MASK	0
-BRD_SER1_RTSCTS	2
+BRD_SER1_RTSCTS	1
```
dat-tran-matternet, line 37: The BBB uart flow control pins (UART5_CTS/RTS) are also used as SYS_BOOT pins for bootup options. These two pins control the selected external crystal frequency:   
    00b = 19.2 MHz
    01b = 24 MHz
    10b = 25 MHz
    11b = 26 MHz

The correct configuration selected by pull up/down resistors is 01b, 24 MHz. However, with the Cube controlling the BBB_UART5_CTS line, there is a good chance it will drive low at the same time BBB boots up. This would cause the boot pins to be 00b, 19.2 MHz. This mismatch in frequency would mess up timing on the BBB. 

We may need the Cube to delay the driving of it's RTS line as a workaround.

Extra notes:  
When a device is ready to accept UART data, it will drive RTS low. Most likely the Cube will boot faster than the BBB and drive the RTS low. This will cause incorrect selection of the crystal frequency  (19.2 MHz).  If the Cube is slower to initialize then RTS would be in high impedance state and allow the correct frequency. This is a race condition.

dat-tran-matternet (approved):

libraries/AP_HAL_ChibiOS/hwdef/MttrCubeOrange/defaults.parm
zalice: @dat-tran-matternet There's an ardupilot parameter change that we plan to test to prevent the Cube from driving those lines for at least the first second. We're tracking that here: https://matternet.atlassian.net/browse/AVX-324

I'll updated the description of that ticket with the detail you shared in this comment.

AydenChubbic (approved):

zalice:
Will add this to firmware defaults when update firmware with all defaults in future cube orange autopilot firmware release.

PR #187, opened 2021-07-14
Copter: Add mttr param tkoff_accel

zalice:
Add TOFF_ACCEL as configurable parameter


zalice:
To be addressed by 4.1

PR #188, opened 2021-07-16
Feature/apm 90/disallow arming on lane 2

Aksellence:
https://matternet.atlassian.net/browse/APM-90
Patch to not allow the aircraft to arm if the aircraft is using an EKF lane other than EKF Lane 1.

### DIFF
--- a/ArduCopter/AP_Arming.cpp
+++ b/ArduCopter/AP_Arming.cpp
@@ -598,12 +598,15 @@ bool AP_Arming_Copter::arm_checks(AP_Arming::Method method)
 
     const NavEKF2 &ekf2 = AP::ahrs_navekf().get_NavEKF2_const();
     const NavEKF3 &ekf3 = AP::ahrs_navekf().get_NavEKF3_const();
-
+    const int ekf_lane1 = 0;
+    const int ekf2_enabled = 2;
+    const int ekf3_enabled = 3;
+    
     // Do not arm if EKF lane 1 is not used as primary during arming, when using a certain EKF (EKF2 or EKF3).
-    if ((ahrs.get_ekf_type() == 2) && (ekf2.currentPrimaryLane() != 0)) {
+    if ((ahrs.get_ekf_type() == ekf2_enabled) && (ekf2.currentPrimaryLane() != ekf_lane1)) {
         check_failed(true, "EKF2 lane 1 not in use as primary");
         return false;
-    } else if ((ahrs.get_ekf_type() == 3) && (ekf3.currentPrimaryLane() != 0)) {
+    } else if ((ahrs.get_ekf_type() == ekf3_enabled) && (ekf3.currentPrimaryLane() != ekf_lane1)) {
         check_failed(true, "EKF3 lane 1 not in use as primary");
         return false;
     }
--- a/ArduCopter/AP_Arming.cpp
+++ b/ArduCopter/AP_Arming.cpp
@@ -596,6 +596,21 @@ bool AP_Arming_Copter::arm_checks(AP_Arming::Method method)
         return false;
     }
 
+    // Do not arm if EKF lane 1 is not used as primary during arming, when using EK2.
+    const NavEKF2 &ekf2 = AP::ahrs_navekf().get_NavEKF2_const();
+    if (ekf2.currentPrimary() != 0) {
+        check_failed(true, "EKF lane 1 not in use as primary");
+        return false;
+    }
+
+    // Do not arm if EKF lane 1 is not used as primary during arming, when using EK3
+    const NavEKF3 &ekf3 = AP::ahrs_navekf().get_NavEKF3_const();
+    if (ekf3.currentPrimary() != 0) {
+        check_failed(true, "EKF lane 1 not in use as primary");
+        return false;
+    }
+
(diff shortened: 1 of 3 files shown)


ArduCopter/AP_Arming.cpp
```diff
@@ -596,6 +596,21 @@ bool AP_Arming_Copter::arm_checks(AP_Arming::Method method)
         return false;
     }
 
+    // Do not arm if EKF lane 1 is not used as primary during arming, when using EK2.
+    const NavEKF2 &ekf2 = AP::ahrs_navekf().get_NavEKF2_const();
+    if (ekf2.currentPrimary() != 0) {
+        check_failed(true, "EKF lane 1 not in use as primary");
```
adriennet823: Suggest saying "EKF2 lane 1" for additional info and same below for EKF3

ArduCopter/AP_Arming.cpp
```diff
@@ -596,6 +596,21 @@ bool AP_Arming_Copter::arm_checks(AP_Arming::Method method)
...
+        return false;
+    }
+
+    // Do not arm if EKF lane 1 is not used as primary during arming, when using EK3
```
zalice: What is the outcome of `ekf3.currentPrimary()` when not using EK3? Should there be a check for which EKX we're using?

libraries/AP_NavEKF2/AP_NavEKF2.h
```diff
@@ -51,6 +51,11 @@ class NavEKF2 {
         return num_cores;
     }
 
+    // Gets the lane number which is used as primary
+    uint8_t currentPrimary(void) const {
```
AydenChubbic: Could you name this function something like`currentPrimaryLane` or `currentPrimaryLaneId`

AydenChubbic (approved):

ArduCopter/AP_Arming.cpp
```diff
@@ -596,6 +596,21 @@ bool AP_Arming_Copter::arm_checks(AP_Arming::Method method)
         return false;
     }
 
+    // Do not arm if EKF lane 1 is not used as primary during arming, when using EK2.
```
bhiln: Should the last word in the comment be `EKF2`?

ArduCopter/AP_Arming.cpp
bhiln: Should comment say `EKF3`?

bhiln: Additionally, is there only one primary? if so, can the second `if` be an `else if`?

ArduCopter/AP_Arming.cpp
Aksellence: That was deliberate as I kept it consistent with how we name the parameters corresponding to each EKF (for example EK2_ENABLE), but yea I will change it to EKFx, rather than EKx to make it more readable.

ArduCopter/AP_Arming.cpp
Aksellence: Same as above.

libraries/AP_NavEKF2/AP_NavEKF2.h
Aksellence: Done. Thanks!

ArduCopter/AP_Arming.cpp
Aksellence: As per the new changes of adding a check with EKF_TYPE, else if makes more sense. Done!

ArduCopter/AP_Arming.cpp
Aksellence: Done

zalice (approved):

AydenChubbic (approved):

ArduCopter/AP_Arming.cpp
```diff
@@ -596,6 +596,18 @@ bool AP_Arming_Copter::arm_checks(AP_Arming::Method method)
         return false;
     }
 
+    const NavEKF2 &ekf2 = AP::ahrs_navekf().get_NavEKF2_const();
+    const NavEKF3 &ekf3 = AP::ahrs_navekf().get_NavEKF3_const();
+
+    // Do not arm if EKF lane 1 is not used as primary during arming, when using a certain EKF (EKF2 or EKF3).
+    if (ahrs.get_ekf_type() == 2 && ekf2.currentPrimaryLane() != 0) {
```
rabad-matternet: Could you add parentheses for better readability?

if ((ahrs.get_ekf_type() == 2) && (ekf2.currentPrimaryLane() != 0))

rabad-matternet: `ahrs.get_ekf_type() == 2`

Can '2' be #defined or made a const expr?

rabad-matternet: `ekf2.currentPrimaryLane() != 0`

Can '0' be a #define or const expr?

bhiln (approved):

ArduCopter/AP_Arming.cpp
Aksellence: Done!

rabad-matternet (approved):

PR #189, opened 2021-07-19
Keep RTS high until data is written to a UART

tridge:
This changes the handling of RTS on telem1 and telem2 so it starts high, and is only activated after the first bytes are written. This combines with TELEM_DELAY to ensure RTS is high during early boot, to prevent an issue with the BBB


rabad-matternet:
After hardware and BBB changes, it was determined that this PR is no longer needed.

PR #190, opened 2021-07-30
AP_ARMING AP_COMPASS limit mag limits furter

zalice:
Tighten the hardcoded mag limits to narrow the mag deviation allowed in prearm checks for the following:
-AP_ARMING_COMPASS_MAGFIELD_MIN
- AP_ARMING_COMPASS_MAGFIELD_MAX
- AP_COMPASS_MAX_XYZ_ANG_DIFF
- AP_COMPASS_MAX_XY_ANG_DIFF
- AP_COMPASS_MAX_XY_LENGTH_DIFF

### DIFF
--- a/libraries/AP_Arming/AP_Arming.cpp
+++ b/libraries/AP_Arming/AP_Arming.cpp
@@ -49,8 +49,8 @@
 #include <AP_Logger/AP_Logger.h>
 
 #define AP_ARMING_COMPASS_MAGFIELD_EXPECTED 530
-#define AP_ARMING_COMPASS_MAGFIELD_MIN  185     // 0.35 * 530 milligauss
-#define AP_ARMING_COMPASS_MAGFIELD_MAX  875     // 1.65 * 530 milligauss
+#define AP_ARMING_COMPASS_MAGFIELD_MIN  200     // 0.35 * 530 milligauss, limited further for mttr ops
+#define AP_ARMING_COMPASS_MAGFIELD_MAX  700     // 1.65 * 530 milligauss, limited further for mttr ops
 #define AP_ARMING_BOARD_VOLTAGE_MAX     5.8f
 #define AP_ARMING_ACCEL_ERROR_THRESHOLD 0.75f
 #define AP_ARMING_AHRS_GPS_ERROR_MAX    10      // accept up to 10m difference between AHRS and GPS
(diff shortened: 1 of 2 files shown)


libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -49,8 +49,8 @@
 #include <AP_Logger/AP_Logger.h>
 
 #define AP_ARMING_COMPASS_MAGFIELD_EXPECTED 530
-#define AP_ARMING_COMPASS_MAGFIELD_MIN  185     // 0.35 * 530 milligauss
-#define AP_ARMING_COMPASS_MAGFIELD_MAX  875     // 1.65 * 530 milligauss
+#define AP_ARMING_COMPASS_MAGFIELD_MIN  200     // 0.35 * 530 milligauss, limited further for mttr ops
+#define AP_ARMING_COMPASS_MAGFIELD_MAX  700     // 1.65 * 530 milligauss, limited further for mttr ops
```
bhiln: Please update the math in the comment to show derivation of new value

libraries/AP_Compass/AP_Compass.h
```diff
@@ -38,9 +38,9 @@
 
 // define default compass calibration fitness and consistency checks
 #define AP_COMPASS_CALIBRATION_FITNESS_DEFAULT 16.0f
-#define AP_COMPASS_MAX_XYZ_ANG_DIFF radians(90.0f)
-#define AP_COMPASS_MAX_XY_ANG_DIFF radians(60.0f)
-#define AP_COMPASS_MAX_XY_LENGTH_DIFF 200.0f
+#define AP_COMPASS_MAX_XYZ_ANG_DIFF radians(30.0f)
+#define AP_COMPASS_MAX_XY_ANG_DIFF radians(20.0f)
+#define AP_COMPASS_MAX_XY_LENGTH_DIFF 100.0f
```
bhiln, line 43: The first two values are 1/3 of the original value, but this one (`AP_COMPASS_MAX_XY_LENGTH_DIFF`) is only 1/2. Is this intentional or is there no correlation to the previous two values?

adriennet823 (approved):
Looks good 👍  Cross-checked this against the proposed changes in #gnc

libraries/AP_Compass/AP_Compass.h
zalice: This is intentional. The first two are angles, the last one is magnitude. Different dimensional planes

bhiln (approved):
Pending one fix, looks great!

libraries/AP_Arming/AP_Arming.cpp
zalice: done

Aksellence (approved):
Looks good.

PR #191, opened 2021-08-11
Copter: implement change alt command in MISSION_ITEM handling

tridge:
this allows for altitude change in missions. Behaviour is:

 - for a changealt command (see cmd_changealt in MAVProxy for an
   example), when in AUTO mode the target mission altitude will change
   for the current waypoint and all future waypoints until it is
   cancel or times out

 - a CHANGEALT_TMOUT parameter is added giving the timeout in
   milliseconds, defaulting to 3000. A value <= 0 means no timeout

 - if a changealt is issued for >= 1km below sea level then that is
   considered a cancel

 - altitudes between 0 and -1000km are ignored

 - only normal waypoints are affected

To test, use the "changealt NNN" command in mavproxy. That command sends a relative-home altitude in meters.


bhiln:
Thanks for putting this together @tridge! I've confirmed that this update integrates into MATE well.

In the description, you mention that "when in AUTO mode the target mission altitude will change for the current waypoint and all future waypoints". the "current waypoint" is the current target mission waypoint, correct? Is it possible to have the aircraft ascend/descend at maximum up/down speed (without forfeiting horizontal speed (within vibration/controls constraints of the aircraft)) until it reaches the target msl alt and then continue at that alt? My goal is to reach the given altitude as soon as possible. It seems that if we set the next mission waypoint alt to target alt, then if the waypoint is very far away, we will slowly reach the target alt, and if it is very close by (horizontally), we will forfeit horizontal speed to reach the adjusted alt fast waypoint before continuing to the next waypoint.

Thanks!

bhiln:
In your description, you mentioned that "altitudes between 0 and -1000km are ignored". If we are flying at a location below sea level and an msl alt of -10 is commanded, will that command be ignored?

bhiln:
Confirmed integration with MATE!
Screen Shot 2021-08-17 at 1 58 56 PM

tridge:
>  Is it possible to have the aircraft ascend/descend at maximum up/down speed (without forfeiting horizontal speed (within vibration/controls constraints of the aircraft)) until it reaches the target msl alt and then continue at that alt?

what limit do we want on vertical speed? Do we want to use the existing WPNAV_SPEED_UP and WPNAV_SPEED_DN parameters?

bhiln:
> > Is it possible to have the aircraft ascend/descend at maximum up/down speed (without forfeiting horizontal speed (within vibration/controls constraints of the aircraft)) until it reaches the target msl alt and then continue at that alt?
> 
> what limit do we want on vertical speed? Do we want to use the existing WPNAV_SPEED_UP and WPNAV_SPEED_DN parameters?

Yep, WPNAV_SPEED_UP and WPNAV_SPEED_DN will be great

tridge:
@bhiln I've updated the PR to climb/descend at WPNAV_SPEED_UP/WPNAV_SPEED_DN as soon as the commanded alt is given. It also fixes it to allow for target altitudes below sea level up to -1000m, anything at -1000m or more is considered a cancel.

tridge:
@bhiln Leonard has fixed the issue with it getting stuck, should be good now

bhiln:
> @bhiln Leonard has fixed the issue with it getting stuck, should be good now

Thank you @tridge and @lthall! I'll be testing today and the first half of this week. I'll post here with anything I find.

bhiln:
@tridge @lthall During my testing, I've found that during an altitude maneuver, if the aircraft brakes, then resumes, the aircraft does not hold the originally commanded altitude - it seems to fly towards the next mission waypoint (in this case a gradual ascend).

Ideally, the aircraft should hold the commanded altitude even during/after a brake/resume, and fly back to the nominal altitude afterwards. (If the aircraft would have been ascending or descending to the next mission waypoint at the given lat/lon, would it make sense to resume flight at the altitude that it was when the changealt was commanded, or the altitude it would have been now if no changealt was commanded?)

In the image below, you can see that the first changealt command completes and then another one begins. After a brake and resume, it looks like the intermediate waypoint is cleared. While it's slowly ascending towards the next mission waypoint, I command another change alt back down and then cancel the command, which resumes the gradual climb.

image

cc: @zalice

bhiln (approved):
We have been testing and things look great now! Thank you for your work @tridge and @lthall !

zalice (approved):

Aksellence (approved):
Looks good.

PR #195, opened 2021-09-07
AC_WPNav: cope with incorrect setting on negative WPNAV_SPEED_DN

tridge:
this prevents a climb away if the parameter is incorrectly set as
negative


bhiln (approved):
Thanks for the protection!

zalice (approved):

PR #196, opened 2021-09-08
AP_Parachute: Update fts temperature logging format

zalice:
Convert and log FTS temperature in degrees Celcius. The temperature is received from FTS in 0.1K

### DIFF
--- a/libraries/AP_Parachute/AP_Parachute.cpp
+++ b/libraries/AP_Parachute/AP_Parachute.cpp
@@ -246,7 +246,7 @@ void AP_Parachute::mttr_fts_update()
                     "FTSS",
                     "TimeUS,State,Rsn,BSoC,BC1mV,BC2mV,BC3mV,BTINT,BTTS1",
                     "s--------",
-                    "----CCCAA",
+                    "-------AA",
                     "QBBBHHHhh",
                     AP_HAL::micros64(),
                     msg->state,
--- a/libraries/AP_Parachute/AP_Parachute.cpp
+++ b/libraries/AP_Parachute/AP_Parachute.cpp
@@ -12,7 +12,7 @@
 
 extern const AP_HAL::HAL& hal;
 
-constexpr float CELCIUS_TO_KELVIN_FACTOR = 273.15;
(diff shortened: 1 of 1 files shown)


Aksellence (approved):
It looks good.

libraries/AP_Parachute/AP_Parachute.cpp
```diff
@@ -240,7 +242,21 @@ void AP_Parachute::mttr_fts_update()
             enum fts_msg_id_t msg_id = fts_protocol_identify_message(msg_len, msg_buf);
             if (msg_id == FTS_MSGID_STATUS) {
                 struct fts_msg_status_s* msg = (struct fts_msg_status_s*)msg_buf;
-                AP::logger().Write("FTSS", "TimeUS,State,Rsn,BSoC,BC1mV,BC2mV,BC3mV,BTINT,BTTS1", "QBBBHHHcc", AP_HAL::micros64(), msg->state, msg->state_reason, msg->batt_SoC, msg->batt_cell_mV[0], msg->batt_cell_mV[1], msg->batt_cell_mV[2], msg->batt_temp_INT, msg->batt_temp_TS1);
+                AP::logger().Write(
+                    "FTSS",
+                    "TimeUS,State,Rsn,BSoC,BC1mV,BC2mV,BC3mV,BTINT,BTTS1",
+                    "s--------",
```
tridge, line 248: the field is called BC1mV, which implies it is is milliVolts, but you've included the 'C' multipler for 1000. That will look odd in a graph for graphing tools that know about the units and multipliers, as the value will be scaled by 0.001 in the tool.
you can also add 'v' in units for volts and 'O' for degrees C so tools will display it correctly. The units string should be "s---vvvOO", and you could remove the mV suffix on columns. Alternatively leave it in millivolts, leave in the mV suffix, and replace the 'C' multipler with 0 meaning no multiplier
Note that the log truncation that is done for some logs loses the units and multiplier annotations

zalice: Ah okay I follow how this works now, thank you! I will leave it in millivolts and remove the multiplier.

PR #197, opened 2021-09-09
Added configurable GPS-Fix prearm check

Aksellence:
- Added configuration options: STATUS_MIN.
- Make GPS status arming check configurable.

### DIFF
--- a/ArduCopter/AP_Arming.cpp
+++ b/ArduCopter/AP_Arming.cpp
@@ -499,6 +499,8 @@ bool AP_Arming_Copter::proximity_checks(bool display_failure) const
 // performs mandatory gps checks.  returns true if passed
 bool AP_Arming_Copter::mandatory_gps_checks(bool display_failure)
 {
+    const AP_GPS &gps = AP::gps();
+
     // always check if inertial nav has started and is ready
     const AP_AHRS_NavEKF &ahrs = AP::ahrs_navekf();
     if (!ahrs.prearm_healthy()) {
@@ -531,12 +533,11 @@ bool AP_Arming_Copter::mandatory_gps_checks(bool display_failure)
         if (reason == nullptr || strlen(reason) == 0) {
             if (!mode_requires_gps && fence_requires_gps) {
                 // clarify to user why they need GPS in non-GPS flight mode
-                reason = "Fence enabled, need 3D Fix";
+                check_failed(display_failure, "Fence enabled, need GPS Fix %u", gps.status_arm_min());
             } else {
-                reason = "Need 3D Fix";
+                check_failed(display_failure, "Needs GPS Fix %u", gps.status_arm_min());
             }
         }
-        check_failed(display_failure, "%s", reason);
         return false;
     }
 
--- a/ArduCopter/AP_Arming.cpp
+++ b/ArduCopter/AP_Arming.cpp
@@ -499,8 +499,6 @@ bool AP_Arming_Copter::proximity_checks(bool display_failure) const
 // performs mandatory gps checks.  returns true if passed
 bool AP_Arming_Copter::mandatory_gps_checks(bool display_failure)
 {
-    const AP_GPS &gps = AP::gps();
-
     // always check if inertial nav has started and is ready
     const AP_AHRS_NavEKF &ahrs = AP::ahrs_navekf();
     if (!ahrs.prearm_healthy()) {
@@ -533,9 +531,9 @@ bool AP_Arming_Copter::mandatory_gps_checks(bool display_failure)
         if (reason == nullptr || strlen(reason) == 0) {
             if (!mode_requires_gps && fence_requires_gps) {
                 // clarify to user why they need GPS in non-GPS flight mode
-                check_failed(display_failure, "Fence enabled, need GPS Fix %u", gps.status_arm_min());
+                check_failed(display_failure, "Fence enabled, need 3D Fix");
             } else {
-                check_failed(display_failure, "Needs GPS Fix %u", gps.status_arm_min());
+                check_failed(display_failure, "Need 3D Fix");
             }
         } else {
             check_failed(display_failure, "%s", reason);
--- a/ArduCopter/AP_Arming.cpp
+++ b/ArduCopter/AP_Arming.cpp
@@ -537,6 +537,8 @@ bool AP_Arming_Copter::mandatory_gps_checks(bool display_failure)
             } else {
                 check_failed(display_failure, "Needs GPS Fix %u", gps.status_arm_min());
             }
+        } else {
+            check_failed(display_failure, "%s", reason);
         }
         return false;
     }
--- a/libraries/AP_GPS/AP_GPS.cpp
+++ b/libraries/AP_GPS/AP_GPS.cpp
@@ -290,13 +290,20 @@ const AP_Param::GroupInfo AP_GPS::var_info[] = {
     AP_GROUPINFO("BLEND_TC", 21, AP_GPS, _blend_tc, 10.0f),
 #endif
 
-
     // @Param: DRV_OPTIONS
     // @DisplayName: driver options
     // @Description: Additional backend specific options
     // @Bitmask: 0:Use UART2 for moving baseline on ublox,1:Use SERIALn_BAUD instead of auto-bauding
     // @User: Advanced
     AP_GROUPINFO("DRV_OPTIONS", 22, AP_GPS, _driver_options, 0),
+
+    // @Param: STATUS_ARM_MIN
+    // @DisplayName: Minimum GNSS fix type required for Arming
+    // @Description: Sets the minimum GNSS fix type needed for arming.
+    // @Values: 0:No GPS,1:NoFix,2:2DFix,3:3DFix,4:SBAS,5:RTKFloat,6:RTK
+    // @User: Advanced
+    // @RebootRequired: True
+    AP_GROUPINFO("STATUS_MIN", 61, AP_GPS, _status_arm_min, 3),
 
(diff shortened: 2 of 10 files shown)


ArduCopter/AP_Arming.cpp
```diff
@@ -531,12 +533,11 @@ bool AP_Arming_Copter::mandatory_gps_checks(bool display_failure)
...
             if (!mode_requires_gps && fence_requires_gps) {
                 // clarify to user why they need GPS in non-GPS flight mode
-                reason = "Fence enabled, need 3D Fix";
+                check_failed(display_failure, "Fence enabled, need GPS Fix %u", gps.status_arm_min());
             } else {
-                reason = "Need 3D Fix";
+                check_failed(display_failure, "Needs GPS Fix %u", gps.status_arm_min());
             }
         }
-        check_failed(display_failure, "%s", reason);
```
zalice, line 539: Could you leave this in so that if reason is not null or empty, it will get called? Wrap it in an else to avoid calling it twice if reason was previously null?

libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -290,14 +290,21 @@ const AP_Param::GroupInfo AP_GPS::var_info[] = {
...
     // @User: Advanced
     AP_GROUPINFO("DRV_OPTIONS", 22, AP_GPS, _driver_options, 0),
 
+    // @Param: STATUS_ARM_MIN
+    // @DisplayName: Minimum GNSS fix type required for Arming
+    // @Description: Sets the minimum GNSS fix type needed for arming.
+    // @Values: 0:No GPS,1:NoFix,2:2DFix,3:3DFix,4:SBAS,5:RTKFloat,6:RTK
+    // @User: Advanced
+    // @RebootRequired: True
+    AP_GROUPINFO("STATUS_MIN", 61, AP_GPS, _status_arm_min, 3),
```
zalice: Could you make this 64? And default SBAS?

libraries/AP_NavEKF2/AP_NavEKF2_Measurements.cpp
```diff
@@ -505,7 +505,7 @@ void NavEKF2_core::readGpsData()
     // do not accept data at a faster rate than 14Hz to avoid overflowing the FIFO buffer
     const AP_GPS &gps = AP::gps();
     if (gps.last_message_time_ms() - lastTimeGpsReceived_ms > 70) {
-        if (gps.status() >= AP_GPS::GPS_OK_FIX_3D) {
+        if (gps.status() >= gps.status_arm_min()) {
```
zalice: We should accept GPS data below the arm min

libraries/AP_NavEKF3/AP_NavEKF3_Measurements.cpp
```diff
@@ -506,7 +506,7 @@ void NavEKF3_core::readGpsData()
     const AP_GPS &gps = AP::gps();
 
     if (gps.last_message_time_ms() - lastTimeGpsReceived_ms > frontend->sensorIntervalMin_ms) {
-        if (gps.status() >= AP_GPS::GPS_OK_FIX_3D) {
+        if (gps.status() >= gps.status_arm_min()) {
```
zalice: same comment here as in the EKF2: we should accept data below GPS status arm min

libraries/AP_GPS/AP_GPS.cpp
Aksellence: 64 is too big of an index and I get an error, so I cannot use 64. Yes I can default to SBAS.

libraries/AP_NavEKF2/AP_NavEKF2_Measurements.cpp
Aksellence: Done

libraries/AP_NavEKF3/AP_NavEKF3_Measurements.cpp
Aksellence: Done

ArduCopter/AP_Arming.cpp
Aksellence: Done

libraries/AP_NavEKF2/AP_NavEKF2_Measurements.cpp
```diff
@@ -642,7 +642,7 @@ void NavEKF2_core::readGpsData()
         } else {
             // report GPS fix status
             gpsCheckStatus.bad_fix = true;
-            hal.util->snprintf(prearm_fail_string, sizeof(prearm_fail_string), "Waiting for 3D fix");
+            hal.util->snprintf(prearm_fail_string, sizeof(prearm_fail_string), "Waiting for GPS fix %u", gps.status_arm_min());
```
zalice: Since we're not requiring enhanced fix, this can be reverted right?

libraries/AP_NavEKF3/AP_NavEKF3_Measurements.cpp
```diff
@@ -654,7 +654,7 @@ void NavEKF3_core::readGpsData()
         } else {
             // report GPS fix status
             gpsCheckStatus.bad_fix = true;
-            hal.util->snprintf(prearm_fail_string, sizeof(prearm_fail_string), "Waiting for 3D fix");
+            hal.util->snprintf(prearm_fail_string, sizeof(prearm_fail_string), "Waiting for GPS fix %u", gps.status_arm_min());
```
zalice: Same comment here about reverting this change

libraries/AP_NavEKF/AP_Nav_Common.h
```diff
@@ -52,7 +52,7 @@ union nav_gps_status {
         bool bad_horiz_drift    : 1; // 5 - true if the GPS horizontal position is drifting (this check assumes vehicle is static)
         bool bad_hdop           : 1; // 6 - true if the reported HDoP is insufficient to start using GPS
         bool bad_vert_vel       : 1; // 7 - true if the GPS vertical speed is too large to start using GPS (this check assumes vehicle is static)
-        bool bad_fix            : 1; // 8 - true if the GPS is not providing a 3D fix
+        bool bad_fix            : 1; // 8 - true if the GPS is not providing the minimum required fix
```
zalice: Bad fix is still 3D (same comment for the other file in which this comment was changed)

libraries/AP_NavEKF2/AP_NavEKF2_Measurements.cpp
Aksellence: Yea you are right. Done.

libraries/AP_NavEKF3/AP_NavEKF3_Measurements.cpp
Aksellence: Done

libraries/AP_NavEKF/AP_Nav_Common.h
Aksellence: Done

libraries/AP_NavEKF2/AP_NavEKF2_Outputs.cpp
```diff
@@ -534,7 +534,7 @@ void  NavEKF2_core::getFilterGpsStatus(nav_gps_status &faults) const
     faults.flags.bad_horiz_drift    = gpsCheckStatus.bad_horiz_drift; // GPS horizontal drift is too large to start using GPS (check assumes vehicle is static)
     faults.flags.bad_hdop           = gpsCheckStatus.bad_hdop; // reported HDoP is too large to start using GPS
     faults.flags.bad_vert_vel       = gpsCheckStatus.bad_vert_vel; // GPS vertical speed is too large to start using GPS (check assumes vehicle is static)
-    faults.flags.bad_fix            = gpsCheckStatus.bad_fix; // The GPS cannot provide the 3D fix required
+    faults.flags.bad_fix            = gpsCheckStatus.bad_fix; // The GPS cannot provide the minimum fix required
```
zalice: Please revert this and AP_NavEKF3_Outputs.cpp

ArduCopter/AP_Arming.cpp
```diff
@@ -531,12 +533,13 @@ bool AP_Arming_Copter::mandatory_gps_checks(bool display_failure)
...
             if (!mode_requires_gps && fence_requires_gps) {
                 // clarify to user why they need GPS in non-GPS flight mode
-                reason = "Fence enabled, need 3D Fix";
+                check_failed(display_failure, "Fence enabled, need GPS Fix %u", gps.status_arm_min());
             } else {
-                reason = "Need 3D Fix";
+                check_failed(display_failure, "Needs GPS Fix %u", gps.status_arm_min());
             }
+        } else {
+            check_failed(display_failure, "%s", reason);
```
zalice: Are these changes needed? I think these still rely on 3D fix, not whatever is configured in gps.status_arm_min(), right?

Aksellence: Given that we are disallowing arming based on the minimum fix needed based on the GPS_STATUS_MIN param I would say we should display the value of 'min fix needed' rather than saying 3D, which could be confusing based on what the param value is set to.

libraries/AP_NavEKF2/AP_NavEKF2_Outputs.cpp
Aksellence: Done

ArduCopter/AP_Arming.cpp
Aksellence: Done

zalice (approved):

zalice (approved):

PR #200, opened 2021-09-19
Support mission pause via DO_CHANGE_SPEED

tridge:
Using DO_CHANGE_SPEED with the speed_type of 0 or 1 now accepts a speed of zero to mean "pause mission". A speed > 0 will unpause mission.
This only pauses the forward part of the mission, so you can combine this with a changealt command to climb/descend while paused.
To test you can use the "setspeed" command in MAVProxy with a speed in m/s


lthall:
Looks good mate!

bhiln (approved):
Looks great and confirmed works as intended in MATE - Thanks @tridge and @lthall !!

adriennet823 (approved):

PR #201, opened 2021-09-19
Fixed a rounding error in magnetic field tables interpolation

tridge:
This fixes an error in the mag field tables that ArduPilot uses to initialise the earth frame field in the EKF at startup. The error could cause significant earth field errors, with the magnitude dependent on the latitude/longitude.
In practice the EKF learned the error fairly quickly, but it is better to start off with the right value
A test suite is added as well to test for correctness


zalice (approved):

Aksellence (approved):
Looks good.

PR #206, opened 2021-10-22
APM-123 Pr report min sat prearm gps2

zalice:
Currently the message for the num sats GPS2 failed prearm check is reported only if GPS1 passes.

This removes the dependency between GPS1 and GPS2 sat count prearm messages.


adriennet823 (approved):
🚀

bhiln (approved):

Aksellence (approved):

PR #210, opened 2021-11-02
APM 126 - Added configurable mag limit Params

Aksellence:
Added configurable mag limit Parameters:
1. **COMPASS_AD_MAX** - Configures Angle difference in XYZ plane (in degrees) 
2. **COMPASS_ADXY_MAX** - Configures Angle difference in XY plane (in degrees)
3. **COMPASS_LDXY_MAX** - Configures Length difference in XY plane (in mG)

### DIFF
--- a/libraries/AP_Compass/AP_Compass.cpp
+++ b/libraries/AP_Compass/AP_Compass.cpp
@@ -645,6 +645,34 @@ const AP_Param::GroupInfo Compass::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("CUS_YAW", 51, Compass, _custom_yaw, 0),
 #endif
+
+    // @Param: AD_MAX
+    // @DisplayName: Compass maximum Angle-difference (AD) in all axes (xyz plane)
+    // @Description: This sets the maximum allowed angle difference in all axes between the two compass' for calibration check and arming consistency checks
+    // @Range: -180 180
+    // @Units: deg
+    // @Increment: 1
+    // @User: Advanced
+    AP_GROUPINFO("AD_MAX", 52, Compass, _ang_diff_xyz_max_deg, 20.0f),
+
+    // @Param: ADXY_MAX
+    // @DisplayName: Compass maximum Angle-difference (AD) in xy plane
+    // @Description: This sets the maximum allowed angle difference in XY axis between the two compass' for calibration check and arming consistency checks
+    // @Range: -180 180
+    // @Units: deg
+    // @Increment: 1
+    // @User: Advanced
+    AP_GROUPINFO("ADXY_MAX", 53, Compass, _ang_diff_xy_max_deg, 10.0f),
+
+    // @Param: LDXY_MAX
+    // @DisplayName: Compass maximum Length-difference (LD) in xy plane
+    // @Description: This sets the maximum allowed length difference in XY axis between the two compass' for calibration check and arming consistency checks
+    // @Range: 0 200
+    // @Units: mGauss
+    // @Increment: 1
+    // @User: Advanced
+    AP_GROUPINFO("LDXY_MAX", 54, Compass, _len_diff_xy_max_mG, 60.0f),
+
     AP_GROUPEND
 };
 
@@ -1906,17 +1934,17 @@ bool Compass::consistent() const
         const float xy_ang_diff  = acosf(constrain_float(mag_field_xy*primary_mag_field_xy_norm,-1.0f,1.0f));
 
         // check for gross misalignment on all axes
-        if (xyz_ang_diff > AP_COMPASS_MAX_XYZ_ANG_DIFF) {
+        if (xyz_ang_diff > radians(_ang_diff_xyz_max_deg)) {
             return false;
         }
 
         // check for an unacceptable angle difference on the xy plane
-        if (xy_ang_diff > AP_COMPASS_MAX_XY_ANG_DIFF) {
+        if (xy_ang_diff > radians(_ang_diff_xy_max_deg)) {
             return false;
         }
 
         // check for an unacceptable length difference on the xy plane
-        if (xy_len_diff > AP_COMPASS_MAX_XY_LENGTH_DIFF) {
+        if (xy_len_diff > _len_diff_xy_max_mG) {
             return false;
         }
     }
--- a/libraries/AP_Compass/AP_Compass.h
+++ b/libraries/AP_Compass/AP_Compass.h
@@ -38,9 +38,6 @@
 
 // define default compass calibration fitness and consistency checks
 #define AP_COMPASS_CALIBRATION_FITNESS_DEFAULT 16.0f
-#define AP_COMPASS_MAX_XYZ_ANG_DIFF radians(20.0f)
-#define AP_COMPASS_MAX_XY_ANG_DIFF radians(10.0f)
-#define AP_COMPASS_MAX_XY_LENGTH_DIFF 60.0f
 
 /**
    maximum number of compass instances available on this platform. If more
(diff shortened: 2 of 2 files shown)


dat-tran-matternet (approved):

adriennet823 (approved):
Thanks @AkshayMatternet !

libraries/AP_Compass/AP_Compass.cpp
```diff
@@ -1906,17 +1946,17 @@ bool Compass::consistent() const
         const float xy_ang_diff  = acosf(constrain_float(mag_field_xy*primary_mag_field_xy_norm,-1.0f,1.0f));
 
         // check for gross misalignment on all axes
-        if (xyz_ang_diff > AP_COMPASS_MAX_XYZ_ANG_DIFF) {
+        if (xyz_ang_diff > radians(get_ang_diff_xyz_max())) {
```
zalice: Instead of calling getter functions here can the class use its private member directly? As it does for other private members in this class?

libraries/AP_Compass/AP_Compass.cpp
```diff
@@ -645,6 +657,34 @@ const AP_Param::GroupInfo Compass::var_info[] = {
...
 #endif
+
+    // @Param: AD_MAX
+    // @DisplayName: Compass maximum Angle-difference (AD) in all axis (xyz plane)
+    // @Description: This sets the maximum allowed angle difference in all axis between the two compass' for calibration check and arming consistency checks
+    // @Range: -180 180
+    // @Units: deg
+    // @Increment: 1
+    // @User: Advanced
+    AP_GROUPINFO("AD_MAX", 52, Compass, _ang_diff_xyz_max, AP_COMPASS_ANG_DIFF_XYZ_MAX),
```
zalice: Instead of predirectives, can the defaults be set directly?

Aksellence: I prefer keeping it consistent with the rest of the file, such that we use predirectives rather than constant if they are defined. Case in point:
- Line 489: AP_GROUPINFO("OFFS_MAX", 31, Compass, _offset_max, AP_COMPASS_OFFSETS_MAX_DEFAULT)
-  Line 510: AP_GROUPINFO("FLTR_RNG", 34, Compass, _filter_range, HAL_COMPASS_FILTER_DEFAULT)
etc.
Although I can add a suffix '_DEFAULT' to make it consistent to better readability and maintain homogeneousity throughout.

zalice: Is it maintaining consistency if there are 55 parameters and 4 have preprocessor directives?
Keeping the default value here prevents the need to scroll up to check the value.

libraries/AP_Compass/AP_Compass.cpp
Aksellence: Yes, that was the plan initially, but I made the getter functions for higher reachability, in case we need to access these values in other classes. And the reason I used the method instead of the variable is to keep it consistent, such that if these values would be accessed in other classes/modules, the same method would be called throughout the code, keeping the accessifiers for these values consistent everywhere.

libraries/AP_Compass/AP_Compass.cpp
Aksellence: By consistency I meant, if there are preprocessor directives, so we should be using them as defaults rather than constants. We can always remove them but seems like the 'Max' value params are set preprocessor directives, so I chose to follow the trend and went with how it was for the 'max param', default values.
I agree that removing the predirectives will make the code smaller and make it more readable, it is simply a choice that was made based on how I saw the code to be, I am more than happy to change it if that is what you think would seem more readable.

libraries/AP_Compass/AP_Compass.h
```diff
@@ -546,6 +558,15 @@ friend class AP_Compass_Backend;
...
     AP_Int16 _offset_max;
 
+    // Angle difference between compass' in xyz plane, in degrees
+    AP_Float _ang_diff_xyz_max;
+
+    // Angle difference between compass' in xy plane, in degrees
+    AP_Float _ang_diff_xy_max;
+
+    // Mag field Length difference between compass' in xy plane, in mG
+    AP_Float _len_diff_xy_max;
```
rabad-matternet: Are these ever updated?  If not, should these be made `const`?

Can you add units (e.g. `_ang_diff_xyz_max_deg`)?

libraries/AP_Compass/AP_Compass.h
```diff
@@ -336,6 +333,21 @@ friend class AP_Compass_Backend;
...
+
+    // return maximum allowed angle difference (in degrees) between compass' in xy plane
+    AP_Float get_ang_diff_xy_max(void) const {
+        return _ang_diff_xy_max;
+    }
+
+    // return maximum allowed mag field length difference (in mG) between compass' in xy plane
+    AP_Float get_len_diff_xy_max(void) const {
+        return _len_diff_xy_max;
+    }
```
rabad-matternet: With C++, the MATE team has adopted the convention that "get" and "set" functions use the same overloaded function name:
```
AP_Float ang_diff_xyz_max(void);          // get function
void ang_diff_xyz_max(AP_Float ang_max);  // set function
```
Should we be applying that convention here as well to have that consistency with MATE?
(@zalice @adriennet823 @AydenChubbic @bhiln @viet-nguyen-matternet if you care to chime on this)

AydenChubbic: If there is no existing nomenclature, I think it makes sense to use the same naming convention across as much of our code base as possible

libraries/AP_Compass/AP_Compass.cpp
zalice: If we need to access this data outside of this class then we can add the accessors. Adding the functions just in case bloats the header.

libraries/AP_Compass/AP_Compass.h
zalice: Good catch on that @rabad-matternet.

More broadly, I'm requesting that the accessor functions be removed since no class other than the this class needs to access this data. If we need it later, we'll add accessors.

libraries/AP_Compass/AP_Compass.cpp
zalice: Since these preprocessor directives are not used elsewhere, most defaults don't have them, and finding the value requires scrolling, I recommend removing them

libraries/AP_Compass/AP_Compass.cpp
Aksellence: I have made the desired changes. Thanks! for the feedback

libraries/AP_Compass/AP_Compass.cpp
Aksellence: I have made the desired changes. Thanks! for the feedback

libraries/AP_Compass/AP_Compass.h
Aksellence: I have made the desired changes. Thanks! for the feedback

viet-nguyen-matternet (approved):

libraries/AP_Compass/AP_Compass.cpp
```diff
@@ -645,6 +645,34 @@ const AP_Param::GroupInfo Compass::var_info[] = {
     // @User: Advanced
     AP_GROUPINFO("CUS_YAW", 51, Compass, _custom_yaw, 0),
 #endif
+
+    // @Param: AD_MAX
+    // @DisplayName: Compass maximum Angle-difference (AD) in all axis (xyz plane)
```
zalice: axis --> axes

libraries/AP_Compass/AP_Compass.cpp
```diff
@@ -645,6 +645,34 @@ const AP_Param::GroupInfo Compass::var_info[] = {
...
+    // @Description: This sets the maximum allowed angle difference in all axis between the two compass' for calibration check and arming consistency checks
```
zalice: axis --> axes

zalice (approved):

PR #211, opened 2021-11-02
APM-121 - Prearm check for GPS2 min allowed status

Aksellence:
Added Prearm check for minimum allowed fix type (status) for GPS2. Previously this check was only applicable for the primary GPS.
https://matternet.atlassian.net/browse/APM-121

### DIFF
--- a/libraries/AP_Arming/AP_Arming.cpp
+++ b/libraries/AP_Arming/AP_Arming.cpp
@@ -433,6 +433,9 @@ bool AP_Arming::gps_checks(bool report)
     const AP_GPS &gps = AP::gps();
     if ((checks_to_perform & ARMING_CHECK_ALL) || (checks_to_perform & ARMING_CHECK_GPS)) {
 
+        bool gps1_status_fail = false;
+        bool gps2_status_fail = false;
+
         //GPS OK?
         if (!AP::ahrs().home_is_set() ||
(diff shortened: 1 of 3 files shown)


dat-tran-matternet (approved):

adriennet823 (approved):
🔥

libraries/AP_Arming/AP_Arming.cpp
```diff
@@ -433,15 +433,27 @@ bool AP_Arming::gps_checks(bool report)
...
         //GPS OK?
         if (!AP::ahrs().home_is_set() ||
             gps.status() < AP_GPS::GPS_OK_FIX_3D) {
             check_failed(ARMING_CHECK_GPS, report, "Bad GPS Position");
             return false;
         }
 
-        if (gps.status() < gps.status_arm_min()) {
-            check_failed(ARMING_CHECK_GPS, report, "Waiting for enhanced GPS fix");
+        if (gps.status(0) < gps.status_arm_min()) {
```
zalice, line 446: If one of the gps modules fail, will this get function return safely?

Aksellence: Yes, It returns a value of 0 if that GPS instance is not active (due to failure or other reasons such as wrong configuration etc).

zalice (approved):

rabad-matternet (approved):

PR #213, opened 2021-12-02
APM-131: Detect compass communication error and reset accordingly

dat-tran-matternet:


viet-nguyen-matternet (approved):

adriennet823 (approved):

viet-nguyen-matternet (approved):

tridge:
@dat-tran-matternet this should work, but is not the right approach for fixing this. The approach we use in ArduPilot for re-init of a sensor after an unexpected power cycle in flight is called "checked registers". You can mark a register as being "checked", and if it changes from the expected value then it is automatically reset.
This wasn't implemented in the AK09916 driver due to some complexity with the auxiliary driver. I've now implemented it for the I2C AK09916 in this PR:
https://github.com/ArduPilot/ardupilot/pull/19571
I've also opened an alternative PR to this one for the mttr-rebase-4.0 branch here:
https://github.com/matternet/ardupilot/pull/215
I don't have an I2C AK09916 handy to test, so I'd appreciate if you could test this

dat-tran-matternet:
A better approach developed by @tridge is here:  https://github.com/matternet/ardupilot/pull/215
Closing this PR

PR #215, opened 2021-12-19
AP_Compass: added register checking for AK09916 mode

tridge:
this allows AK09916 on I2C to recover from a power outage in flight
This is an alternative to #213 that uses the checked register API

### DIFF
--- a/libraries/AP_Compass/AP_Compass_AK09916.cpp
+++ b/libraries/AP_Compass/AP_Compass_AK09916.cpp
@@ -210,6 +210,9 @@ bool AP_Compass_AK09916::init()
         goto fail;
     }
 
+    // one checked register for mode
+    _bus->setup_checked_registers(1);
+
     if (!_setup_mode()) {
         hal.console->printf("AK09916: Could not setup mode\n");
         goto fail;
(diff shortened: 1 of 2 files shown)


dat-tran-matternet:
The i2c-mag is not able to recover based on this test procedure:

Test Procedure:

1. Power on AP Board with Cube Orange and Mag connected
2. Connect Cube’s USB to computer
3. Run "mavproxy.py --master /dev/ttyACM0 --console --load-module=graph"
4. Run "graph SCALED_IMU3.zmag"
5. Rotate Mag and verify signal corresponds to rotation
6. Unplug the Mag and verify signal is flat
7. Re-plug the Mag and verify signal corresponds to rotation

The test fails at step 7. thanks

tridge:
@dat-tran-matternet my apologies, I lost part of the fix when I moved this PR from ardupilot master to mttr-rebase-4.0. I didn't think I had an I2C AK09916 to test with, but I've now found one, and I've fixed the PR and tested it. It continues correctly after removing the AK009916 from the bus and plugging it back in

dat-tran-matternet (approved):

dat-tran-matternet:
> @dat-tran-matternet my apologies, I lost part of the fix when I moved this PR from ardupilot master to mttr-rebase-4.0. I didn't think I had an I2C AK09916 to test with, but I've now found one, and I've fixed the PR and tested it. It continues correctly after removing the AK009916 from the bus and plugging it back in

Tested on my end as well and test passed. thanks!

zalice (approved):

PR #219, opened 2022-01-12
APM-134: Secondary Mag health

dat-tran-matternet:


dat-tran-matternet:
Preliminary review. Not tested yet.

libraries/AP_Compass/AP_Compass.h
```diff
@@ -183,7 +183,7 @@ friend class AP_Compass_Backend;
 
     /// Return the health of a compass
     bool healthy(uint8_t i) const { return _get_state(Priority(i)).healthy; }
-    bool healthy(void) const { return healthy(0); }
```
zalice, line 186: Is this function used in contexts that check whether the primary compass is healthy? If so, would it make sense to avoid touching that logic and write an `all_healthy()` function instead?

zalice: Actually, looking at the code again now I don't think this change to `healthy()` is needed because the healthy mask that gets reported in the named value int you added doesn't use this, right?

dat-tran-matternet: This function "compass.healthy()" is used in arming check and in the system status sensor_health. If any one compass is down it will be 0/false.  At the same time, a new NAMED_INT_VALUE is sent COMPHEALTH which is also a bitmask where the bit is the health of each compass.

dat-tran-matternet: Here are all the places that uses "compass.healthy()"
Do we require all of those use cases to require all compass to be healthy? That was my assumption with this PR.

https://github.com/matternet/ardupilot/blob/mttr-rebase-4.0/ArduCopter/AP_Arming.cpp#L620
https://github.com/matternet/ardupilot/blob/mttr-rebase-4.0/ArduCopter/Log.cpp#L268
https://github.com/matternet/ardupilot/blob/mttr-rebase-4.0/libraries/AP_Arming/AP_Arming.cpp#L394
https://github.com/matternet/ardupilot/blob/mttr-rebase-4.0/libraries/GCS_MAVLink/GCS.cpp#L137

zalice: @dat-tran-matternet You're right -- we require all compasses to be healthy in each of those cases. Thank you for clarifying 👍

zalice (approved):

PR #231, opened 2022-02-15
APM-145: Add motor kill option for test

dat-tran-matternet:


ArduCopter/motors.cpp
```diff
@@ -7,6 +7,10 @@
 
 static uint32_t auto_disarm_begin;
 
+#define MOTOR_COUNT  4
+static bool motor_kill_all;
+static bool motor_kill[MOTOR_COUNT];
```
rabad-matternet: Just a suggestion....

Could you use a bit mask for motor_kill?  Then motor_kill = 0xF would be equivalent to motor_kill_all

rabad-matternet (approved):

ArduCopter/motors.cpp
dat-tran-matternet: I was going to do this but noticed that Tridge's implementation for rebase-3.x.x had an explicit flag for kill all. The subtle difference is that the 2nd dimension provided by motor_kill_all flag allows one to remember the motor states if kill_all is switched off. I don't know the use case for this test so I'm inclined to copy what was done prior.

Reference: https://github.com/matternet/ardupilot/pull/136/files

zalice (approved):

PR #236, opened 2022-05-28
Extended GNSS fix loss debug

zalice:
Add GCS error messages for GNSS fix loss

### DIFF
--- a/libraries/AP_GPS/AP_GPS_UBLOX.cpp
+++ b/libraries/AP_GPS/AP_GPS_UBLOX.cpp
@@ -1266,6 +1266,8 @@ AP_GPS_UBLOX::_parse_gps(void)
         }else{
             next_fix = AP_GPS::NO_FIX;
             state.status = AP_GPS::NO_FIX;
+            GCS_SEND_TEXT(MAV_SEVERITY_ERROR,
+                            "Unexpected GNSS fix status");
         }
 #if UBLOX_FAKE_3DLOCK
         state.status = AP_GPS::GPS_OK_FIX_3D;
@@ -1283,6 +1285,7 @@ AP_GPS_UBLOX::_parse_gps(void)
         state.hdop = 170;
 #endif
         break;
+    // Not used by F9P
     case MSG_SOL:
         Debug("MSG_SOL fix_status=%u fix_type=%u",
               _buffer.solution.fix_status,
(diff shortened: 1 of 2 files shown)


libraries/AP_GPS/AP_GPS_UBLOX.cpp
```diff
@@ -1405,6 +1408,18 @@ AP_GPS_UBLOX::_parse_gps(void)
         state.location.lng    = _buffer.pvt.lon;
         state.location.lat    = _buffer.pvt.lat;
         state.location.alt    = _buffer.pvt.h_msl / 10;
+        // Check if valid fix
+        if (!(_buffer.pvt.flags & 0b00000001)
```
adriennet823, line 1412: Is there a doc defining which flag correlates to which bit within the bitmap that I can reference?

adriennet823 (approved):

libraries/AP_GPS/AP_GPS_UBLOX.cpp
```diff
@@ -1405,6 +1408,18 @@ AP_GPS_UBLOX::_parse_gps(void)
         state.location.lng    = _buffer.pvt.lon;
         state.location.lat    = _buffer.pvt.lat;
         state.location.alt    = _buffer.pvt.h_msl / 10;
+        // Check if valid fix
+        if (!(_buffer.pvt.flags & 0b00000001) 
+            || (_buffer.pvt.fix_type != 3 && _buffer.pvt.fix_type != 4)) {
```
viet-nguyen-matternet, line 1413: the bitmasks `0b00000001`, `3` and `4` would help to not be magic numbers to provide clarity.

viet-nguyen-matternet (approved):

libraries/AP_GPS/AP_GPS_UBLOX.cpp
zalice: https://cdn.sparkfun.com/assets/learn_tutorials/1/1/7/2/ZED-F9R_Interfacedescription__UBX-19056845_.pdf

libraries/AP_GPS/AP_GPS_UBLOX.cpp
zalice: I'd recommend we submit a refactor to the main ardupilot branch (if this clarify is important) so that our changes do not diverge more than necessary and add complexity to our rebase to future ardupilot updates.

PR #238, opened 2022-07-18
APM-163: Add degraded GPS functionality

robkarol:


libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -873,14 +873,48 @@ void AP_GPS::update(void)
...
+
 /*
   update primary GPS instance
  */
 void AP_GPS::update_primary(void)
 {
 #if defined(GPS_BLENDED_INSTANCE)
+    uint8_t num_failed = update_and_count_degraded_gps();
+
+    if(num_failed > 0 && GPS_MAX_RECEIVERS > 1) {
```
adriennet823: nitpick: please add a space between `if` and `(`. Same goes for elsewhere throughout this PR. Thank you!

libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -873,14 +873,48 @@ void AP_GPS::update(void)
...
 
+/*
+    determines if any gps's are degraded
+*/
+uint8_t AP_GPS::update_and_count_degraded_gps(void) {
+    uint8_t count = 0;
+    for (uint8_t i = 0; i < GPS_MAX_RECEIVERS; i++) {
+        if(state[i].status < GPS_OK_FIX_3D) {
+            _degraded_gps[i] = true;
+            count++;
```
adriennet823: recommend `++count` for performance

libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -905,13 +939,13 @@ void AP_GPS::update_primary(void)
         return;
     }
 
-    if (_auto_switch == 0) {
+    if (_auto_switch == GPS_AUTO_SWITCH_USEFIRST) {
```
adriennet823, line 943: Nice, I like the use of the enum 👍

libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -873,14 +873,48 @@ void AP_GPS::update(void)
...
+        // If the second GPS failed, use the first.
+        if(_degraded_gps[1]) {
+            _auto_switch = GPS_AUTO_SWITCH_USEFIRST;
+        }
+    }
+
+    // All gps's have failed at some point, use the best moving forward
+    if(num_failed == GPS_MAX_RECEIVERS) {
+        _auto_switch = GPS_AUTO_SWITCH_USEBEST;
+    }
```
adriennet823: Should this be evaluated first? With the current logic, if both GPSs have failed, it looks like we'll only check if GPS1 failed first, change `_auto_switch` to use second, then check if GPS2 also failed, which then changes `_auto_switch` to use first, then we check if both (all) GPSs have failed/degraded to < 3D fix and only then switch to use best.

adriennet823 (changes requested):
Left some questions on order of evaluation of degraded GPSs

libraries/AP_GPS/AP_GPS.cpp
bhiln: `i` in the for loop as well

libraries/AP_GPS/AP_GPS.cpp
bhiln: +1

libraries/AP_GPS/AP_GPS.h
```diff
@@ -521,6 +521,16 @@ class AP_GPS
...
     AP_Int8 _status_arm_min;
 
+    bool _degraded_gps[GPS_MAX_RECEIVERS]; // Any GPS that has previously failed is considered degraded
+
+    // Auto_Switch
+    enum GPS_AUTO_SWITCH {
+        GPS_AUTO_SWITCH_USEFIRST = 0, // Effectively USEFIRST
+        GPS_AUTO_SWITCH_USEBEST = 1,
+        GPS_AUTO_SWITCH_BLEND = 2,
+        GPS_AUTO_SWITCH_USESECOND = 3
```
bhiln, line 531: Is there a reason why `USESECOND` is not 1, next to `USEFIRST`?

bhiln (approved):

libraries/AP_GPS/AP_GPS.cpp
robkarol: I'll swap, but for built in types I don't think there is a performance difference between pre and post increment.

I choose ++ to stick with ardupilot style but you are right thats probably not important.

libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -873,14 +873,48 @@ void AP_GPS::update(void)
...
 }
 
+/*
+    determines if any gps's are degraded
+*/
+uint8_t AP_GPS::update_and_count_degraded_gps(void) {
+    uint8_t count = 0;
+    for (uint8_t i = 0; i < GPS_MAX_RECEIVERS; i++) {
+        if(state[i].status < GPS_OK_FIX_3D) {
+            _degraded_gps[i] = true;
```
adriennet823: Just to clarify, the degraded GPS array tracks if a GPS has had any sort of degradation or failure to below 3D fix so we'll never re-blend because we never clear these slots, right? Just want to make sure that I understand the logic to be: 1. never go back to blending 2. only change primary GPS if the initial primary GPS from the first GPS failure also experiences a failure

libraries/AP_GPS/AP_GPS.h
robkarol: Yep. I left the constants the same as what Ardupilot had prior to my change.

Once I added the ENUM, this isn't as big an issue since I already broke compatibility with Ardupilot, but I think it might help prevent long term confusion by keeping the state numbers the same.

libraries/AP_GPS/AP_GPS.cpp
robkarol: Yep. this is intentional. As you noted it is increasing the number of changes of _auto_switch (and the number of computations) it prevents more complicated logic for reading. Ex. multiple nested if statements. I'll take another pass to see if nested logic makes it easier to read.

libraries/AP_GPS/AP_GPS.cpp
robkarol: Yep! If the GPS becomes degraded (falls below 3D fix) it never clears. You make a good point though that its probably clearer if I explicitly initialize the array to false at the start.

One clarification though. I am not changing what is primary but rather the _auto_switch toggle. The difference between these is that ardupilot logic has many places where the primary GPS might change, but _auto_switch allows for a long term preference of what the primary GPS gets set to (prevents ardupilot from swapping back due to other conditions).

libraries/AP_GPS/AP_GPS.cpp
adriennet823: Got it, no worries

libraries/AP_GPS/AP_GPS.cpp
adriennet823: Sounds good

libraries/AP_GPS/AP_GPS.cpp
adriennet823: Understood. That makes sense. Thank you for clarifying

adriennet823 (approved):
Can you please change the name of the PR to: "APM-163: Add basic degraded GPS functionality"? So we have the nice auto-traceability b/w git and Jira 🙏 Thank you!

PR #242, opened 2022-07-26
rkarol/degraded gps

robkarol:


adriennet823 (approved):

libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -907,9 +907,12 @@ uint8_t AP_GPS::update_and_count_degraded_gps(void) {
     for (uint8_t i = 0; i < GPS_MAX_RECEIVERS; ++i) {
         if (state[i].status < GPS_OK_FIX_3D) {
             _degraded_gps[i] = true;
+        }
+        if (_degraded_gps[i]) {
```
adriennet823, line 911: Why do we check again to increment the counter?

robkarol: What we were doing before was incrementing the counter in the above if statement (line 909). The problem with that was it was only counting the GPSs which didn't have a 3D fix at the moment which this function was called.

In the new if statement layout. We are now looping through the whole degraded gps array to count up which ones are set to True.

The order makes sure that _degraded_gps gets updated before it gets counted. And counting separately makes sure we include the gps's degraded previously.

adriennet823: Thanks for explaining as this wasn't immediately obvious to me

PR #247, opened 2023-01-11
AS-206 Fix SNR reporting for RFND messaging

rabad-matternet:
- fixed type format character for SNR in RFND message

Before: (in-flight data)
image (3)

After: (Bench testing with M22013)
image (4)


adriennet823 (approved):
Great find! Thanks for your help testing. Do you mind if I update the description and Jira ticket with screenshots of the before and after SNR data?

bhiln (approved):

PR #248, opened 2023-01-12
AS-183 Remote ID emergency status broadcast

bhiln:
Ported from [Copter-4.0-ODID](https://github.com/ArduPilot/ardupilot/tree/Copter-4.0-ODID) [EMERGENCY broadcast commit](https://github.com/ArduPilot/ardupilot/commit/4485ba0dfa5cbe197492688b7706949d987611c2).

Broadcast RID EMERGENCY status:
* set emergency status if chute is released
* if we are armed with no GPS fix and we haven't specifically allowed for non-GPS operation then declare an emergency
* if we are disarmed and falling at over 3m/s then declare an emergency. This covers cases such as deliberate crash with advanced failsafe and an unintended reboot or in-flight disarm
* if we have watchdogged while armed then declare an emergency

`pkt_statustext` variable is removed as it is not used in the code and does not exist in the Ardupilot `Copter-4.0-ODID` branch.

Confirmed EMERGENCY status broadcast upon parachute deploy in SITL.

### DIFF
--- a/libraries/AP_OpenDroneID/AP_OpenDroneID.cpp
+++ b/libraries/AP_OpenDroneID/AP_OpenDroneID.cpp
@@ -271,7 +271,10 @@ void AP_OpenDroneID::send_location_message()
     if (parachute != nullptr && parachute->released()) {
         uav_status = MAV_ODID_STATUS_EMERGENCY;
     }
-    fprintf(stdout, "HAL_PARACHUTE_ENABLED, released: %s, uav_status: %s\n", parachute->released(), uav_status);
+    if (parachute != nullptr) {
+        fprintf(stdout, "parachute released: %s\n", parachute->released());
+    }
+    fprintf(stdout, "HAL_PARACHUTE_ENABLED, uav_status: %s\n", uav_status);
 #endif
     // Not yet implemented in Ardupilot 7.0.4
     // if (AP::vehicle()->is_crashed()) {
--- a/libraries/AP_OpenDroneID/AP_OpenDroneID.cpp
+++ b/libraries/AP_OpenDroneID/AP_OpenDroneID.cpp
@@ -42,8 +42,6 @@
 #include <AP_AHRS/AP_AHRS.h>
 #include <AP_Parachute/AP_Parachute.h>
 #include <AP_Vehicle/AP_Vehicle.h>
-#include <stdio.h>
-#include <unistd.h>
 
 extern const AP_HAL::HAL &hal;
 
@@ -271,10 +269,6 @@ void AP_OpenDroneID::send_location_message()
     if (parachute != nullptr && parachute->released()) {
         uav_status = MAV_ODID_STATUS_EMERGENCY;
     }
-    if (parachute != nullptr) {
-        fprintf(stdout, "parachute released: %s\n", parachute->released());
-    }
-    fprintf(stdout, "HAL_PARACHUTE_ENABLED, uav_status: %s\n", uav_status);
 #endif
     // Not yet implemented in Ardupilot 7.0.4
     // if (AP::vehicle()->is_crashed()) {
--- a/libraries/AP_OpenDroneID/AP_OpenDroneID.cpp
+++ b/libraries/AP_OpenDroneID/AP_OpenDroneID.cpp
@@ -261,13 +261,15 @@ void AP_OpenDroneID::send_location_message()
     if (!ahrs.get_position(current_location)) {
         return;
     }
-        uint8_t uav_status = hal.util->get_soft_armed()? MAV_ODID_STATUS_AIRBORNE : MAV_ODID_STATUS_GROUND;
+    
+    uint8_t uav_status = hal.util->get_soft_armed()? MAV_ODID_STATUS_AIRBORNE : MAV_ODID_STATUS_GROUND;
 #if HAL_PARACHUTE_ENABLED
     // set emergency status if chute is released
     const auto *parachute = AP::parachute();
     if (parachute != nullptr && parachute->released()) {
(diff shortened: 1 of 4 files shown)


adriennet823 (approved):
🆘 📣

libraries/AP_OpenDroneID/AP_OpenDroneID.cpp
```diff
@@ -259,7 +260,20 @@ void AP_OpenDroneID::send_location_message()
...
+    
+    uint8_t uav_status = hal.util->get_soft_armed()? MAV_ODID_STATUS_AIRBORNE : MAV_ODID_STATUS_GROUND;
+#if HAL_PARACHUTE_ENABLED
+    // set emergency status if chute is released
+    const auto *parachute = AP::parachute();
+    if (parachute != nullptr && parachute->released()) {
+        uav_status = MAV_ODID_STATUS_EMERGENCY;
+    }
+#endif
+    // Not yet implemented in Ardupilot 7.0.4
```
robkarol: Can you add a comment to this code on why its being added commented out.

Ex. This should be uncommented once ... happens.

And add to that comment a TODO: AS-XXXX to track the work for when it does get done?

bhiln: will do

bhiln: N/A with merge from [Copter-4.0-ODID](https://github.com/ArduPilot/ardupilot/compare/Copter-4.0-ODID)

robkarol:
Approving based on the code but please add a comment to the PR talking about the second action taken in the PR.

In addition to porting over the changes referenced this PR removes pkt_statustext

robkarol (approved):

bhiln:
> Approving based on the code but please add a comment to the PR talking about the second action taken in the PR.
> 
> In addition to porting over the changes referenced this PR removes pkt_statustext

@robkarol `pkt_statustext` was removed in this PR because it looks like it was missed in the last port. The variable is not used anywhere. I'll make a note of this in the description.

libraries/AP_OpenDroneID/AP_OpenDroneID.cpp
```diff
@@ -254,12 +256,36 @@ void AP_OpenDroneID::send_location_message()
...
+    // if we are armed with no GPS fix and we haven't specifically
+    // allowed for non-GPS operation then declare an emergency
+    if (got_bad_gps_fix && armed && !option_enabled(Options::AllowNonGPSPosition)) {
+        uav_status = MAV_ODID_STATUS_EMERGENCY;
+    }
+
+    // if we are disarmed and falling at over 3m/s then declare an
+    // emergency. This covers cases such as deliberate crash with
+    // advanced failsafe and an unintended reboot or in-flight disarm
+    if (!got_bad_gps_fix && !armed && gps.velocity().z > 3.0) {
```
adriennet823, line 281: `WPNAV_SPEED_DN` looks like it can be configured from 100 to 500 cm/s based on ArduPilot allowed range. We configure it to 225 cm/s (2.5 cm/s) so this velocity _should_  theoretically only reach 2.25m/s. However, if we configure it to 500 m/s, then wouldn't a standard landing meet the criteria and trigger an emergency declaration?

adriennet823: I wonder if instead of hard-coding 3 m/s we should use `gps.velocity().z > max(wpnav_speed_dn, land_speed_high)`

adriennet823 (commented):
Question about gps.velocity().z check

libraries/AP_OpenDroneID/AP_OpenDroneID.cpp
bhiln: One of the other criteria is that the aircraft is not armed. Therefore, this should not trigger during a standard landing.

adriennet823: Oh thank you, I missed that. My brain read that as `armed` 😅

bhiln: In regards to using `max(wpnav_speed_dn, land_speed_high)` instead of hard coding 3, I think that this calculation needs to consider the fact that the aircraft is falling rather than landing. I don't think that the descent rate with props off and parachute deployed should be compared to a controlled landing speed. However, it would be worth looking into determining what a good threshold for our aircraft is and having the ability to set an AP param with this speed to compare to

adriennet823: I agree. I used those parameters as an example because I missed the fact that this could not be a controlled landing descent. Those parameters don't make sense for the situation anymore. Do we know where 3 came from? @tridge is there a parameter we can use in place of the hard-coded value that would make sense?

bhiln:
Merging this PR to unblock flight test. If any changes are needed to address freefall, we can make a new PR.

libraries/AP_OpenDroneID/AP_OpenDroneID.cpp
tridge: note that this threshold is only used when disarmed, so won't trigger on normal flight. It is just there to prevent triggering when on the ground

tridge: @bhiln we certainly could set a param, or perhaps just this:
```
#ifndef ODID_FREEFALL_THRESHOLD
#define ODID_FREEFALL_THRESHOLD 3.0
# endif
````
and set it in hwdef.dat?

PR #249, opened 2023-01-19
AS-215 version in build artifact file name

bhiln:
Add AP version, latest mttr git tag in commit history, and git commit to `.apj` and `.elf` build artifact files.
image

### DIFF
--- a/Tools/scripts/mttr-build-ci.sh
+++ b/Tools/scripts/mttr-build-ci.sh
@@ -26,8 +26,20 @@ echo $PATH
 ./waf configure --board=sitl --debug --enable-opendroneid
 ./waf copter
 
-mv build/MttrCubeBlack/bin/arducopter.apj "/tmp/deploy_files/copter-MttrCubeBlack.apj"
-mv build/MttrCubeOrange/bin/arducopter.apj "/tmp/deploy_files/copter-MttrCubeOrange.apj"
-mv build/MttrCubeBlack/bin/arducopter "/tmp/deploy_files/copter-MttrCubeBlack.elf"
-mv build/MttrCubeOrange/bin/arducopter "/tmp/deploy_files/copter-MttrCubeOrange.elf"
+# Get Ardupilot version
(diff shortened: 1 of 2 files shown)


bhiln:
Check hash, this may be the wrong hash

Tools/scripts/mttr-build-ci.sh
```diff
@@ -26,8 +26,21 @@ echo $PATH
...
-mv build/MttrCubeOrange/bin/arducopter "/tmp/deploy_files/copter-MttrCubeOrange.elf"
+# Get Ardupilot version
+eval $(sed -n 's/^#define  *\([^ ]*\)  *\(.*\) *$/export \1=\2/p' ArduCopter/version.h)
+
+# Get mttr git tag (replace all dashes with underscores)
+git fetch --tags --no-recurse-submodules
+MTTR_GIT_TAG=$(echo $(git tag -l "mttr-*" --sort=-committerdate | head -1) | tr - _)
+
+# Get mttr git hash
+MTTR_GIT_HASH=$(git rev-parse --short=7 HEAD)
```
robkarol, line 37: When this gets run is very important. It must be run after the commit gets merged to the main for this to work. If you can run a test that shows the hash is correct we can leave this in. Otherwise it will cause way more confusion than its worth.

robkarol:
This change may not be needed depending on a few conversations I have later this week. Lets hold off on merging it for now, also pending the test I suggested in the review.

Tools/scripts/mttr-build-ci.sh
bhiln: You are right. I ensured that it worked well on my pc, but I'm not sure it was right from circleci. I'm looking into this.

bhiln: @robkarol Here's the latest push with correct hash in the files: https://github.com/matternet/ardupilot/actions/runs/4008770308
image
image

PR #258, opened 2024-11-14
Feat/AS-822 Set alt target z and desired velocity z in BRAKE mode

adriennet823:
Removes check of `pos_control->is_active_z()` before setting altitude target z and desired velocity z when switching to BRAKE mode. Now, regardless of the value of `pos_control->is_active_z()`, the alt target and velocity z will be set.

Background: https://matternet.atlassian.net/browse/AS-815

Jira ticket: https://matternet.atlassian.net/browse/AS-821

Note: The only change is removing the check of `is_active_z()` in `ModeBrake::init()`. The rest of the changes are clang formatting.

### DIFF
--- a/ArduCopter/mode_brake.cpp
+++ b/ArduCopter/mode_brake.cpp
@@ -7,90 +7,89 @@
  */
 
 // brake_init - initialise brake controller
-bool ModeBrake::init(bool ignore_checks)
-{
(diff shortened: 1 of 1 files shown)


adriennet823:
Closing because clang formatting issues make it hard to see true code diff. See #259

PR #259, opened 2024-11-14
Quick Stop Stage 1

robkarol:
When brake mode initializes it always sets the current state and velocity to the current measured values.

Background: https://matternet.atlassian.net/browse/AS-815

Jira ticket: https://matternet.atlassian.net/browse/AS-821


adriennet823 (approved):
Looks good, thanks!

rabad-matternet (approved):

PR #262, opened 2025-09-02
Fix/AS-989: Change blended to single GPS switch logic

adriennet823:
Currently, if a GPS does not update within 400ms, it is possible for the autopilot to erroneously select this stale GPS as primary, if the stale GPS has a higher number of satellites or equal fix to the blended solution. The logic should validate the last updated time, status, and number of satellites first before switching away from blended to the desired GPS.

Compare to https://github.com/ArduPilot/ardupilot/blob/1ba82e4c07aff8d80875cc4717052068b17a709f/libraries/AP_GPS/AP_GPS.cpp#L1155-L1190

Jira ticket: https://matternet.atlassian.net/browse/AS-989
NCT with occurrence: https://matternet.atlassian.net/browse/NCT-3299

### DIFF
--- a/libraries/AP_GPS/AP_GPS.cpp
+++ b/libraries/AP_GPS/AP_GPS.cpp
@@ -1025,19 +1025,33 @@ void AP_GPS::update_primary(void)
     if (primary_instance == GPS_BLENDED_INSTANCE) {
         primary_instance = 0;
         for (uint8_t i=1; i<GPS_MAX_RECEIVERS; i++) {
-            // choose GPS with highest state or higher number of satellites with at least
-            if (state[i].status < GPS_OK_FIX_3D) {
+            // Choose GPS with highest state or higher number of satellites.
+            // Reject a GPS with an old update time, as it may be the old timestamp that triggered the loss of blending
+            const uint32_t delay_threshold_ms = 400;
+            const bool higher_status = state[i].status > state[primary_instance].status;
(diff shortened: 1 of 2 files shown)


libraries/AP_GPS/AP_GPS.cpp
```diff
@@ -1025,19 +1025,33 @@ void AP_GPS::update_primary(void)
...
+            // Select the new GPS if the primary has old data,
+            // or new GPS either has higher status,
+            // or has the same status and more satellites
+            if ((old_data_primary && !old_data) || higher_status || (equal_status && more_sats)) {
                 primary_instance = i;
-                _last_instance_swap_ms = now;
             }
         }
-        gcs().send_text(MAV_SEVERITY_CRITICAL, "GPS Switch: Switched from Blended to %u", primary_instance+1);
+        _last_instance_swap_ms = now;
```
adriennet823, line 1054: Moved this down here to match the ArduPilot master branch. Would consider moving it back up into the if stmt above if that's how we want the logic to run

adriennet823: Also removed the alert message. This should be discussed first...do we want the alert? Is it useful?
Are there other ways to detect if a GPS stopped sending messages in-time (within 400ms)?

rabad-matternet (approved):

